PR #102, opened 2026-07-22
feat: add floating desktop pet (opt-in, static position, size control)

byunjuneseok:
## Summary

Adds an opt-in **floating desktop pet**: the current companion Pokémon (or egg) floats over the desktop in a borderless transparent panel.

- **Never moves on its own** — no wandering logic. The sprite only animates in place (Gen-V GIF); the user repositions it by dragging. The origin is persisted in UserDefaults and restored on relaunch, falling back to the bottom-right corner of the main screen when the saved position is off-screen (monitor layout change).
- **Settings → Floating Pet group** — enable toggle (default **off**) and a size slider (48–192 px, step 8, default 96). Both persisted via the existing UserDefaults pattern in `UsageStore`; en/ko/ja strings added.

## Implementation

- New `FloatingPetController` (`UI/FloatingPetPanel.swift`): non-activating borderless `NSPanel` at `.floating` level, visible on all Spaces (`canJoinAllSpaces` + `fullScreenAuxiliary`), `isMovableByWindowBackground` as the only way it moves. Enable/size changes are observed with the `withObservationTracking` re-register pattern (same as `AppDelegate.observeStore`); species/shiny changes propagate through SwiftUI environment observation inside the panel.
- **Energy discipline** (per the status-item idle-CPU rules): when disabled or on display sleep, the hosting content view is torn down (same pattern as `popoverDidClose`) so the SwiftUI tree and the GIF frame loop fully stop; it is rebuilt on wake/re-enable. Wired into `AppDelegate.setDisplayAwake`.
- Reuses `SpriteView` as-is (static-first load, animated GIF with shiny fallback, egg glyph fallback) — no new sprite plumbing.

## Testing

- `swift test`: 283 tests passed (3 skipped), including new `testFloatingPetSettingsDefaultAndPersistence` covering defaults (off / 96 px) and persistence across a store reload.
- Manually verified on the installed app: panel appears bottom-right at 96×96 on the floating window layer (checked via CGWindowList), drag-to-reposition persists, toggle/size react live from Settings.

### DIFF
--- a/Sources/PokeTokenBar/Core/Localization.swift
+++ b/Sources/PokeTokenBar/Core/Localization.swift
@@ -106,6 +106,16 @@ struct L {
     var todayCost: String { t("오늘 비용 ($)", "Today's cost ($)", "本日のコスト ($)") }
     var limitPercent: String { t("한도 %", "Limit %", "上限 %") }
     var allOffHint: String { t("전부 끄면 캐릭터만 표시됩니다", "All off shows only the character", "すべてオフにするとキャラクターのみ表示") }
+    // MARK: 플로팅 펫
+    var floatingPetSectionTitle: String { t("플로팅 펫", "Floating Pet", "フローティングペット") }
+    var floatingPetEnableLabel: String { t("플로팅 펫 표시", "Show floating pet", "フローティングペットを表示") }
+    var floatingPetHint: String {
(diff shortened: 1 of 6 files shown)


chattymin:

PR #105, opened 2026-07-22
Fix active Pokédex entries and scope provider status

coin-seeker:
## Summary

- show the current active companion in the Pokédex immediately after hatching
- keep the persisted dex graduation-only so evolution updates do not create duplicates and Fresh Egg semantics stay unchanged
- read the Claude Code and Codex API status components instead of company-wide status indicators
- update English, Korean, and Japanese documentation

## Root cause

The collection view only read permanently graduated entries even though its empty-state copy promised that a hatch would appear there. Provider health checks also consumed the global OpenAI/Anthropic indicator, allowing unrelated product incidents such as Image Generation degradation to appear as a Codex outage.

## Testing

- `./scripts/test-gate.sh` — 288 tests passed, 3 opt-in tests skipped, 84.74% core line coverage
- `swift build -c release`
- regression coverage for loaded active-companion state, hatch/evolve/graduate deduplication, Fresh Egg behavior, and unrelated global incidents


chattymin:
Nice fix — surfacing the active companion in the Pokédex and scoping the status check to the actual component both look solid, and the regression tests reproduce the exact trigger conditions. 👍

Two small follow-up ideas for the active (in-progress) entry, both optional and non-blocking:

1. **Pin the active companion to the top.**
   The synthetic entry has `caughtAt == nil`, so `dexEntriesSorted` treats it as `.distantPast` and it sinks to the bottom of its rarity tier.

   Since it's the one you're actively raising, it reads better pinned above the graduated entries — e.g. keep it out of the rarity/date sort and prepend it rather than sorting it in.

2. **Add a "Raising" badge.**
   The active row currently looks almost identical to a graduated one (same rarity badge + "done" chain); only the missing caught-date sets it apart.

   A small "Raising" pill on that row would make it clear it hasn't graduated yet — and would keep a 3-stage line shown at stage 1 (a single sprite) from being mistaken for a collected single-form.

   Would need en/ko/ja strings like the other labels.

Could both land as a quick follow-up PR. @coin-seeker

coin-seeker:
Thanks for the suggestions — I went ahead and folded both into this PR in `76ed639`.

The active companion is now kept out of the rarity/date sort and prepended above graduated entries, and the row gets a localized `Raising` / `키우는 중` / `育成中` pill. The badge is scoped to the synthetic active entry rather than `caughtAt == nil`, so legacy graduated entries without a date are not mislabeled.

I also added regression coverage for the ordering and all three labels. `./scripts/test-gate.sh` passes with 290 tests (3 opt-in tests skipped, 84.73% core line coverage), and `swift build -c release` passes.

chattymin:

PR #112, opened 2026-07-23
fix: weekly total under-counts in the first days of each month

pangil5634:
## Problem

The weekly / monthly / 5-hour-block "enrichment" pass for the Claude, Codex, and
Gemini providers scans usage logs with a single mtime lower bound of
`startOfMonth(now)`. Usage is derived from append-only session logs, so a file
whose mtime falls before that bound is assumed to contain no in-range entries.

That assumption breaks at the **start of a month**: the current ISO week — and,
just after midnight, the current 5-hour block — can begin in the *previous* month,
so `weekStart < monthStart`. Session files last written in the previous month are
then skipped by the `mtime >= monthStart` filter, and their tokens are dropped from
the "this week" total and the burn-rate block, even though those days belong to the
current week.

In 2026, 11 of 12 months have a week that straddles the month boundary, so the
weekly total silently under-counts for the first few days of nearly every month.
(Being mid-month, the bug is dormant right now — it next triggers at the start of
the following month, which is why it is easy to miss.)

### Root cause

The scan-window computation was duplicated across providers and drifted. The
OpenCode/Hermes provider (added more recently) already guarded this with
`min(startOfMonth, now - 7 days)`, but the three original providers
(Claude / Codex / Gemini) were never updated.

## Fix

Introduce one shared helper, `LocalUsageReader.enrichmentScanStart(now:)`, returning
the earliest of the three windows the scan must cover:

```
min(startOfMonth(now), startOfWeek(now), now - blockWindow)
```

All four providers now route through this single definition, so the window can no
longer drift per provider. Mid-month it equals `startOfMonth` (no extra scanning);
at a month boundary it reaches back to `weekStart` / the block start as needed.

## Tests

- `LocalUsageReaderTests.testEnrichmentScanStartCoversAllWindows` — pure: the bound is
  ≤ each of the month / week / block starts, and equals `startOfMonth` mid-month.
- `LocalUsageCacheTests.testEnrichmentScanStartCatchesWeekStraddlingMonthBoundary` —
  exercises the real mtime filter: with a month-boundary `now`, a session dated in the
  current week but the previous month is dropped under the old `monthStart` bound and
  retained under `enrichmentScanStart`, and lands in the weekly total.

`swift test` — all 298 tests pass (5 skipped).

### DIFF
--- a/Sources/PokeTokenBar/Core/LocalAdditionalUsageProvider.swift
+++ b/Sources/PokeTokenBar/Core/LocalAdditionalUsageProvider.swift
@@ -79,7 +79,9 @@ private actor LocalAdditionalUsageCache {
         }
         if let task = inFlight[source] { return await task.value }
 
-        let periodStart = min(LocalUsageReader.startOfMonth(now), now.addingTimeInterval(-7 * 86400))
+        // 스캔 하한은 enrichment 의 세 윈도우(블록·주·월) 중 가장 이른 시작 — 월초 경계 흡수.
+        // (Claude/Codex/Gemini 경로와 이 단일 소스를 공유해 과거의 window 드리프트를 원천 차단.)
+        let periodStart = LocalUsageReader.enrichmentScanStart(now: now)
         // OpenCode messages are immutable. After the cold monthly read, only query
         // records at and after the newest cached timestamp (with a small overlap).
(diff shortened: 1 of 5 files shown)


chattymin:

PR #118, opened 2026-07-24
fix: branching evolution lines overflow the popover and clip its content

### LINKED ISSUE #117: Evolution line: branching lines (Eevee) overflow the popover width and clip all content
## Summary

While raising a Pokémon whose evolution line branches — Eevee being the extreme case — the
popover content renders wider than the popover's fixed frame and is clipped on **both** edges.
The damage is not limited to the evolution line: the tab bar, the "today's tokens" figures, the
provider row and the limit rows are all cut off on the left and the right.
Image

## Steps to reproduce

1. Raise an Eevee (species 133), i.e. any state where the active Pokémon has not evolved yet.
2. Open the menu-bar popover on the Home tab.

**Expected** — the evolution line stays inside the popover, and the rest of the popover is
unaffected.

**Actual** — the whole popover content is shifted and clipped on both sides; text such as
`Claude Code` and the today's-total figure is sliced off mid-glyph.

## Root cause

`EvoLineView` (`Sources/PokeTokenBar/UI/CompanionView.swift`) lays its nodes out in an `HStack`
with no width constraint, and `CompanionStore.lineNodes` returns *the current path plus every
next candidate*. A branching line is therefore far longer than a linear one:

- Eevee at stage 0 → 1 node + 8 candidates (Vaporeon … Sylveon) = **9 nodes**.
- Measured row width at the default 40pt thumbnail is `54N − 14`, so **N = 9 → 472pt**, against a
  popover content box of `360 − 14×2 = 332pt`. Overflow already starts at **7 nodes**.

`PopoverView` pins the popover with `.frame(width: 360)`, and that modifier *centers* an oversized
child rather than shrinking it. The excess (~70pt per side) therefore hangs outside the popover
window on both sides, where the window clips it — which is why rows that have nothing to do with
the evolution line are cut off as well.

Linear lines are 2–3 nodes (≤ 148pt) and stay well inside the box, which is why this went
unnoticed until a branching line was raised.

## Environment

- PokeTokenBar 2.4.5
- macOS 14+

kimjunseong04:
## Summary

Raising a Pokémon whose evolution line branches makes the popover render wider than itself, and the
whole popover content — tab bar, today's totals, provider row, limit rows — is clipped on **both**
edges, not just the evolution line. This constrains the line to the popover width and scrolls it
horizontally instead. Closes #117.

### Root cause

`EvoLineView` lays its nodes out in an `HStack` with no width constraint, and
`CompanionStore.lineNodes` returns *the current path plus every next candidate*. A branching line is
therefore far longer than a linear one:

- Eevee at stage 0 → 1 node + 8 candidates (Vaporeon … Sylveon) = **9 nodes**.
- Measured row width at the default 40pt thumbnail is `54N − 14`, so **N = 9 → 472pt**, against a
  popover content box of `360 − 14×2 = 332pt`. Overflow already starts at **7 nodes**.

`PopoverView` pins the popover with `.frame(width: 360)`, and that modifier *centers* an oversized
child instead of shrinking it. The excess (~70pt per side) hangs outside the popover window on both
sides, where the window clips it — which is why unrelated rows are cut off too. Linear lines are 2–3
nodes (≤ 148pt) and stay well inside the box, so this never showed up before a branching line was
raised.

### Fix

`EvoLineView` takes a `maxWidth` and scrolls horizontally once the row exceeds it, **keeping the
40pt thumbnails**. Rows that fit take the previous code path unchanged, so ordinary 2–3 stage lines
and every Pokédex row render exactly as before.

Scaling the thumbnails down to fit was the alternative. It was rejected because fitting 9 nodes into
332pt drops them to ~27pt, which makes the sprites hard to tell apart at exactly the moment they
matter most — choosing between eight Eevee evolutions — and it would make sprite size vary with
branch count, so the same species would look different depending on where it sits in its line.

The available width cannot be measured with a `GeometryReader`: the popover width is fixed and an
overflowing child inflates its parent, so the measurement would be circular. `PopoverMetrics` now
holds the popover width and padding as the single source that `PopoverView` and its children share.

Scrollability has to be discoverable, so:

- **Edge fade** on whichever side still has content — it reads as "continues", not "cut off".
- **Chevron button** on that same side, which doubles as the affordance and as the control for mouse
  users. It pages from the current offset with a `.leading` anchor; a fixed target with a
  `.trailing` anchor stopped short of the end and then did nothing on a second press.
- **Native indicator disabled.** With the "always show scroll bars" setting macOS draws a thick
  legacy scroller that also steals row height. `.hidden` does not suppress that — only `.never` does.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

Only the evolution line changes, and only when it does not fit.

| Before | After |
| --- | --- |
| before_clean | after_clean | 
Content bleeds past both edges and is clipped by the window — the `홈` tab, the leading digits of the today's total and the `Claude Code` label are all cut mid-glyph. | The line is bounded and scrolls at unchanged 40pt thumbnails; every other row is fully visible. The trailing edge fades under a `›` button. |

Pressing `›` lands exactly at the end of the line, where the button flips to `‹` and the last
evolution is fully visible:

- **Linear lines (2–3 nodes)** are untouched — no scroll container, no fade, no buttons, and the same
  rendered width as before (asserted by a test).
- **Pokédex rows** are unchanged; their chains are at most 3 nodes and never reach the threshold.

## Testing

`EvoLineLayoutTests` measures rendered width through `NSHostingController.sizeThatFits(in:)`:

- `testEeveeLineWithoutMaxWidthOverflowsPopoverContent` — reproduces the trigger, so the fit
  assertion below cannot pass vacuously.
- `testEeveeLineFitsPopoverContentWidth` / `testLongDexChainFitsCardWidth` — the bounded line stays
  within the popover and the Pokédex card.
- `testShortChainRendersInlineAtNaturalWidth`, `testRowWidthMatchesRenderedWidth` — the width formula
  agrees with the real layout, and short lines keep their previous width.
- `testNeedsScrollBoundaryForHomeLine` / `...ForDexRow`, `testAffordanceShowsOnlyWhereContentRemains`,
  `testAffordanceHiddenWhenContentFits`, `testPageTargetAdvancesAndStaysInBounds` — pure coverage of
  the scroll threshold, the affordance direction at start / middle / end, and chevron paging bounds.

`swift build` clean; `swift test` — all 315 tests pass (5 skipped). Also verified in a real build by
raising an Eevee: the popover is no longer clipped, one chevron press lands exactly at the end of
the line and the forward chevron disappears, one press back returns to the start.

## Checklist

### DIFF
--- a/Sources/PokeTokenBar/UI/CompanionView.swift
+++ b/Sources/PokeTokenBar/UI/CompanionView.swift
@@ -135,17 +135,176 @@ struct SpriteView: View {
 }
 
 /// 진화 라인(초기→최종, 다음 후보 미리보기). done/cur/future.
+///
+/// 분기 라인은 "현재 경로 + 다음 후보 전부"라 길다(이브이 = 본체 1 + 후보 8 → 40pt 기준 472pt).
+/// 폭 제한 없는 HStack 은 팝오버 콘텐츠 폭(332pt)을 넘고, **넘친 자식이 부모 VStack 폭을 부풀려
+/// 팝오버 전체가 좌우로 잘린다**(진화줄뿐 아니라 탭바·합계까지). `maxWidth` 를 주면 그 폭 안에서
+/// 가로 스크롤한다 — 썸네일 크기는 유지하고, 가장자리 페이드 + 셰브론으로 스크롤 가능함을 알린다.
 struct EvoLineView: View {
     let nodes: [(id: Int, kind: String)]
     var thumb: CGFloat = 40
     var shiny: Bool = false     // 개체가 shiny 면 라인 전체를 shiny 스프라이트로
     var names: [Int: String]? = nil   // 제공되면 각 스프라이트 밑에 작은 이름 라벨(도감 단계별 이름)
+    /// 한 줄이 쓸 수 있는 가로 폭. 기본 .infinity = 제한 없음(스크롤 없이 나열).
+    var maxWidth: CGFloat = .infinity
+
+    private static let spacing: CGFloat = 2
+    /// 화살표 칸 폭 = 썸네일 × 이 비율. 고정 frame 을 줘 SF Symbol 글리프 폭에 의존하지 않게 한다 —
+    /// rowWidth 가 실제 렌더 폭과 어긋나면 스크롤 판정이 틀어진다.
+    private static let arrowRatio: CGFloat = 0.25
+    /// 이름 라벨이 썸네일보다 넓어질 수 있는 최대치.
+    private static let nameSlack: CGFloat = 6
+    private static let fadeWidth: CGFloat = 24
+
+    @State private var scrollX: CGFloat = 0        // 현재 가로 스크롤 오프셋
+    @State private var contentWidth: CGFloat = 0   // 실제 렌더된 한 줄 폭(측정값)
+
+    /// 한 줄이 차지하는 가로 폭. 레이아웃과 같은 식을 쓰는 순수 함수 — 이름 라벨은 상한만 알 수
+    /// 있어(`.frame(maxWidth:)`) names 가 있으면 실제 폭이 이 값 이하일 수 있다.
+    static func rowWidth(count: Int, thumb: CGFloat, hasNames: Bool) -> CGFloat {
+        guard count > 0 else { return 0 }
+        let column = thumb + (hasNames ? nameSlack : 0)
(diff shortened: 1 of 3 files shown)


chattymin:

PR #120, opened 2026-07-25
Feat: add floating pet interactivity

Borisserz:
## Summary

Fixes #110. Makes the floating pet interactive: hover usage callout, click to open the popover, drag to move, right-click context menu — on a non-activating LSUIElement panel.

### Review follow-ups (addressed)

1. **Context menu targets** — `target = self`, `autoenablesItems = false`, items force-enabled; menu popped via `rightMouseDown` + `NSMenu.popUpContextMenu` (NSHostingView does not reliably call `menu(for:)`).
2. **Click vs drag** — background-move disabled; drag via `mouseDragged` + `setFrameOrigin`; click uses a pure movement threshold.
3. **Localization** — menu + callout strings through `L` (ko/en/ja). Hide renamed to **Turn off floating pet** (persists until Settings).
4. **Compact utilization gate** — `highestLimitUtilization` only includes providers with today tokens > 0; Codex `$` spend / `individualLimit` excluded.
5. **Hover** — custom callout panel via `NSTrackingArea(.activeAlways)` (NSToolTip rarely appears while inactive).
6. **README ko/ja** + PR template filled.
7. **Review media** — removed from `assets/pr-120/` (owner request); hosted on a temporary release instead of git blobs.

## Type of change

- [x] New feature
- [x] Bug fix (review follow-ups)

## UI changes

| Before | After |
| ------ | ----- |
| Pet is drag-only | Hover → today's tokens callout |
| | Click → opens Token Bar popover |
| | Drag → moves without opening |
| | Right-click → Open Token Bar / Turn off floating pet (enabled + clickable) |

### Recording / screenshots

Full clip (hover → click → drag → right-click):  
https://github.com/Borisserz/PokeTokenBar/releases/download/pr-120-review-evidence/hover-click-drag-menu.mp4

| Hover | Right-click menu | Click opens popover |
| ----- | ---------------- | ------------------- |
| hover | menu | click |

Release tag (temporary review evidence, not product assets): https://github.com/Borisserz/PokeTokenBar/releases/tag/pr-120-review-evidence

## Checklist

- [x] `swift build` and `swift test` pass locally
- [x] PR title and description are written in English
- [x] UI changes are described above (before/after — images optional)
- [x] No copyrighted assets, secrets, or private tooling references are committed (see [CONTRIBUTING](../CONTRIBUTING.md))
- [x] Tests were added or updated for this change

PR #124, opened 2026-07-26
Floating pet: show a localized speech bubble when a usage-limit alert fires (#109)

### LINKED ISSUE #109: Floating pet: show a localized speech bubble when a usage-limit alert fires
## Summary

When the floating desktop pet is enabled and visible, surface "approaching limit" alerts on the pet as a transient, auto-dismissing speech bubble — in addition to the existing system notification.

## Why

The limit-alert engine and its localized copy already exist; the floating pet (added in #102) is currently a passive sprite with no reactions. Showing the alert on the pet makes it more glanceable, without adding a second alert path.

## Approach

- **Reuse the existing alert engine.** Drive the bubble from `UsageStore.evaluateLimitAlerts` — the same edge-triggered warn/crit output that already sends notifications (`UsageStore.swift`). Do **not** add a second threshold check; keep a single source of truth so the bubble and the notification always fire on the same edge.
- **Localized copy.** Reuse the existing localization layer (`store.l`, e.g. `notifWarning` / `notifBody`, ko/en/ja). Add short bubble-specific strings if the notification copy is too long for a bubble.
- **Presentation.** Render the bubble in `FloatingPetPanel` above the pet sprite; auto-dismiss after a few seconds; never steal focus (the panel is a non-activating `NSPanel`).

## Behavior (case table)

| Floating pet | Alert tier | Behavior |
|---|---|---|
| Off | warn / crit | Notification only (unchanged) |
| On, visible | warn | Notification + bubble (auto-dismiss) |
| On, visible | crit | Notification + bubble (crit styling / sound per existing policy) |
| On, display asleep / pet hidden | any | No bubble (notification only) |
| Multiple windows over threshold | — | Show highest severity only (no stacking) |
| Low Power Mode | warn / crit | Bubble shown but static (no animation) |

## Energy

The bubble is transient — tear down its view tree after dismiss (same discipline as `popoverDidClose` / the pet's `hide()`), no persistent relayout on the always-on-top surface.

## Testing

Keep the "alert → bubble copy / severity" mapping a pure function, separate from the AppKit presentation (mirrors the notification path). The alert decision itself is already covered by `evaluateLimitAlerts` tests.

## Open decisions

- Auto-dismiss duration (suggest ~6s fixed).
- Reuse `notifBody` copy vs. add a shorter pet-specific string set (suggest short pet-specific).

Borisserz:
## Summary

Fixes #109. When the floating pet is enabled, surface limit alerts as a transient speech bubble on the pet — **independently** of Notification Center.

### Review follow-ups (addressed)

1. **Decouple from system notifications.** Bubbles no longer ride inside `guard limitNotifications`. Gate is `floatingPetEnabled && floatingPetBubbleAlerts` only.
2. **New setting** (default on), nested under Floating Pet below the size slider: **Limit alerts as bubbles** (`floatingPetBubbleAlerts`, ko/en/ja).
3. **Panel geometry.** `targetFrame` grows for bubble headroom/width; sprite is bottom-anchored; drag persistence stores the *pet* origin so the pet does not jump when a bubble appears.
4. **Japanese horizontal clipping.** Dropped intrinsic horizontal `.fixedSize`; text column is `bubbleMinWidth − 16` with wrap; `bubbleHeadroom` raised to 72. Covered by `testLocalizedAlertBubbleFitsDefaultPanel` (ko/en/ja).
5. **Pure alert → bubble mapping.** `checkLimitAlerts()` evaluates once and advances `notifiedTier` once above both gates; `bubbleAlert(from:)` and `shouldDismissBubble` are pure and unit-tested.
6. **Merged #120 interactivity** (hover/click/drag/menu) into this branch so it can land after #120.
7. **README ko/ja** synced; **review media** removed from `assets/pr-124/` and hosted on a temporary release.

### Note on multi-alert refreshes

One refresh can produce several alerts: Notification Center still posts N notifications; the pet shows a single bubble (highest severity / utilization). Stacking is deferred.

## Type of change

- [x] New feature
- [x] Bug fix (review follow-ups on the bubble path)

## UI changes

| Before | After |
| ------ | ----- |
| Pet is a passive sprite | Limit warn/crit shows a localized speech bubble above the pet for ~6s |
| Mute Notification Center also killed the bubble | Bubble has its own toggle under Floating Pet (default on) |
| Square `size×size` panel clipped the bubble | Panel grows for the bubble; pet origin stays fixed |
| Japanese body could clip horizontally | Text wraps inside a fixed-width column fitting the 180pt panel |

### Screenshots / recording

| Before | After |
| ------ | ----- |
| before | after |

Appear → dismiss: https://github.com/Borisserz/PokeTokenBar/releases/download/pr-124-review-evidence/bubble-appear-dismiss.gif

Release tag (temporary review evidence): https://github.com/Borisserz/PokeTokenBar/releases/tag/pr-124-review-evidence

## Checklist


chattymin:
Two decisions on my side first — #2 adds a setting, so please read before reworking.

## Decisions

**1. Decouple the bubble from system notifications.** It currently rides inside
`checkLimitNotifications()`, which opens with `guard limitNotifications else { return }`
(`UsageStore.swift:771`) — so muting Notification Center also kills the bubble. That's
backwards: the in-app bubble is the lighter channel and the natural substitute for
someone who muted notifications.

**2. New setting — bubble alerts on/off, nested under the floating pet. Default on.**

```
Floating Pet
  ├─ Show floating pet            [toggle]
  └─ (when enabled)
      ├─ Size                     [slider]   ← existing
      └─ Limit alerts as bubbles  [toggle]   ← new, default on
```

- `UsageStore`: `var floatingPetBubbleAlerts: Bool`, same `didSet` shape as
  `floatingPetEnabled` (`:75-77`), read back as
  `d.object(forKey: "floatingPetBubbleAlerts") as? Bool ?? true` (mirrors `:345`).
- `SettingsView.floatingPetGroup`: a `toggleRow` inside the existing
  `if store.floatingPetEnabled { … }` block (`SettingsView.swift:170-178`), below the
  size slider — same nesting as the notifications group at `:187`.
- `Localization.swift`: ko/en/ja label next to `floatingPetSizeLabel` (`:117`).

The gate becomes `floatingPetEnabled && floatingPetBubbleAlerts`, with no reference to
`limitNotifications`.

## 1. Blocker — the bubble can't render at any pet size

The `contentRect` change from 96×96 to 200×200 (`FloatingPetPanel.swift:115`) is
overridden one call later:

```
show()             → p.setFrame(targetFrame(size: floatingPetSize), display: true)   :81
targetFrame(size:) → NSSize(width: size, height: size)                               :101
floatingPetSize    → default 96, slider 48…192   (UsageStore.swift:346)
```

The window is a `size × size` square, but `VStack(spacing: 8) { bubble; sprite }` needs
`bubbleHeight + 8 + size` of height and — via `.fixedSize(horizontal: true, …)` — a
bubble wider than `size`. Content outside a window's frame isn't composited, so the
bubble is clipped away. At the default 96px the window would need ~144px; raising the
slider doesn't help because the window stays square.

The fix belongs in `targetFrame` plus the drag-persistence math: reserve headroom and
width, bottom-anchor the sprite, and keep `windowDidMove` (`:134`) / `defaultOrigin`
(`:109`) storing the *pet's* origin — otherwise the pet jumps whenever a bubble appears.

Please attach a recording of the bubble appearing and dismissing.

## 2. The alert→bubble decision needs to be a pure function

Issue #109 asked for this:

> **Testing** — Keep the "alert → bubble copy / severity" mapping a pure function,
> separate from the AppKit presentation (mirrors the notification path).

As written the block sits after `guard AppEnv.isBundledApp else { return }` (`:772`),
which is false under xctest — nothing past it can run in a test, which is why this PR has
none. `UsageStore.swift` is also in the coverage gate's `LOGIC_CORE` set
(`scripts/test-gate.sh:24`).

Decision (1) needs the same refactor, so they collapse into one change:

```swift
private func checkLimitAlerts() {
    let windows = buildLimitWindows()                      // extracted from :776-812
    let alerts = Self.evaluateLimitAlerts(                 // tier advances once, centrally
        windows: windows, warn: warnThreshold, crit: critThreshold, tiers: &notifiedTier)
    guard !alerts.isEmpty else { return }
    if limitNotifications, AppEnv.isBundledApp { postLimitNotifications(alerts) }
    if floatingPetEnabled, floatingPetBubbleAlerts { showBubble(Self.bubbleAlert(from: alerts)) }
}
```

Two things to watch:

- `notifiedTier` must advance **exactly once, above both gates** — not per consumer.
  Advancing it in each branch would desync the two channels and re-fire suppressed alerts
  (the #61-era regression this state exists to prevent).
- This changes behavior intentionally: today the tier never advances while notifications
  are off, so re-enabling them fires a burst. After the refactor re-enabling is quiet. I
  think that's better — flagging it so it's a decision, not a side effect.

`Self.bubbleAlert(from:)` and the 6s expiry check are then pure. Please test both.

## 3. README ko/ja

Only `README.md` got the new bullet; the repo keeps all three in sync.

## Nice to have

- `sorted` isn't a stable sort, so with two critical windows (Claude weekly + Codex
  primary is a common pair) the displayed alert is unspecified, and utilization is
  ignored within a tier. `alerts.max { ($0.isCritical, $0.utilization) < ($1.isCritical, $1.utilization) }`.
- One refresh can produce several alerts; N notifications post but only one bubble shows.
  Fine to defer — just note it in the description.
- Two description claims don't match the code: "tearing down the view tree" (setting
  `currentBubbleAlert = nil` removes one subview; real teardown is `hide()`'s
  `p.contentView = nil` at `:88`) and "obeys Low Power Mode" (only the animation is
  skipped).
- `currentBubbleAlert` is `public var` — everything else on this class is unannotated or
  `private(set)`.
- `L(store.localizationLanguage)` works, but that's the notification mirror seed
  (`PokeTokenBarApp.swift:46`); `FloatingPetView` already holds
  `@Environment(CompanionStore.self)`.

---

Please fill in the PR template checklist. Use PR Template.
@Borisserz

Borisserz:
## Bubble recording / screenshots (review request)

Real captures from a running build with Floating Pet + Limit alerts as bubbles enabled:

### Before (pet only)
before

### After (bubble visible — clipped correctly, not cut off by the panel)
after

### Appear → dismiss (~6s cycle)
appear-dismiss

Files also live in `assets/pr-124/` on this branch.

chattymin:
Both decisions landed exactly as described — the gate is `floatingPetEnabled && floatingPetBubbleAlerts` with no reference to `limitNotifications`, and the new setting sits where I sketched it with all three locales. The `evaluateLimitAlerts` / `bubbleAlert` split is the right shape and the three tests cover it.

The vertical half of the blocker is genuinely fixed: the panel grows upward, the sprite is bottom-anchored, and `panelSize` / `panelOrigin` / `petOrigin` are pure with a round-trip test. Thanks for the recording.

The horizontal half is still open.

## 1. The bubble is clipped in Japanese

`SpeechBubbleView` still ends with `.fixedSize(horizontal: true, vertical: false)`, so the bubble takes its intrinsic width, while the panel width is the constant `max(petSize, bubbleMinWidth)` = 180 at the default pet size. Anything wider is outside the window frame and isn't composited — the same failure as before, just on the other axis.

Measured with `NSAttributedString` against the real alert strings (`Localization.swift:317-321`), title at 11pt semibold, body at 10pt, plus the 10pt horizontal padding on each side:

| Locale | Body string | Bubble width | Result |
|---|---|---|---|
| ko | Claude 5시간 세션 한도 85% 사용 | 165.9pt | fits (14.1pt spare) |
| en | Claude 5-hour session at 85% | 166.3pt | fits (13.7pt spare) |
| **ja** | **Claude 5時間セッション 上限 85% 使用** | **188.9pt** | **clipped by 8.9pt** |
| ko/ja | Codex 주간 / 週間 … | 135.6pt | fits |

It only disappears if the size slider is pushed past 189, which isn't a fix.

Two related things while you're in there:

- Vertical headroom measures 53.9pt against `bubbleHeadroom = 56` — it passes, but with 2.1pt of slack. One more word or a metrics change and it clips.
- `.lineLimit(2)` is unreachable today: with `fixedSize(horizontal: true)` the text never wraps, so the bubble is always one line.

## Suggested fix

I'd fix the width by constraining it rather than by measuring it — measuring means keeping an AppKit measurement in step with SwiftUI's rendering, and a small mismatch clips again silently.

1. Drop `.fixedSize(horizontal: true, vertical: false)`.
2. Give the bubble a fixed content width, e.g. `.frame(width: FloatingPetController.bubbleMinWidth - 16, alignment: .leading)`.
3. Raise `bubbleHeadroom` from 56 to ~72 so two wrapped lines fit with slack.

That also makes `.lineLimit(2)` meaningful. If you'd rather keep the bubble hugging its text, measuring the string inside `panelSize` works too — it stays pure and testable — but please add padding for the metric mismatch.

## Please lock the Japanese case with a test

`testPanelGrowsForBubbleWithoutChangingPetOrigin` compares constants, so it passes today even though ja clips — that's the false confidence I'd like to remove. A test that runs the actual localized strings for ko/en/ja and asserts the rendered bubble fits inside `panelSize(petSize: 96, showingBubble: true)` would have caught this, and would catch the next string change too.

## Also: PR evidence in the repo

`assets/pr-124/` adds 642 KB (`bubble-after.png` 346 KB, `bubble-before.png` 144 KB, `bubble-appear-dismiss.gif` 152 KB) to the history permanently — squash-merging doesn't drop the blobs. Could you `git rm -r assets/pr-124` and attach the same files to the PR description instead? `assets/` is for product documentation like the README screenshots.

I'd like to merge this one last: #120 rewrites the drag handling in the same file, so doing the width fix on top of that means reconciling the frame math once rather than twice.
@Borisserz

Borisserz:
Addressed the remaining review items:

1. **Japanese horizontal clipping** — content width capped to `bubbleMinWidth − 16`, horizontal padding 8pt/side, `bubbleHeadroom = 72`, wrap instead of intrinsic growth. New fit test covers ko/en/ja (`testLocalizedAlertBubbleFitsDefaultPanel`).
2. **Review media** — removed `assets/pr-124/` from the branch; screenshots/GIF on a temporary release: https://github.com/Borisserz/PokeTokenBar/releases/tag/pr-124-review-evidence
3. **Merged #120** interactivity into this branch so merge order #120 → #124 stays clean.

Please merge #120 first, then this one.

chattymin:
Merged — thank you for your contribution! 🙇‍♂️

The frame rework in particular (bottom-anchored sprite, pure `panelSize`/`panelOrigin`/`petOrigin` with a round-trip test) and the pure alert→bubble mapping made this easy to verify.

This also lands the work from #120, which I'm closing as superseded.

Issue #115, opened 2026-07-24
Add support for a new AI coding agent (usage provider) — architecture & contribution guide


chattymin:
## What this is

PokeTokenBar tracks the tokens you burn across AI coding CLIs. As new agents keep
appearing, adding one should stay a small, well-contained change. This is a
**standing invitation** to add a new usage source (any AI CLI / agent that writes
local usage data) and the guide for doing it without bending the app's architecture.

Want to add a specific agent? Either:
- **Open a new issue that references this one** (e.g. `Add <agent> usage tracking`), so each agent can be tracked separately, **or**
- **Open a PR directly** that follows the checklist below and links back here.

## The one extension point

A usage source is a single implementation of the `UsageProvider` protocol, registered
in one array. **These are the only two places you should need to touch:**

1. **Implement `UsageProvider`** — `Sources/PokeTokenBar/Core/UsageProvider.swift:4`
   ```swift
   protocol UsageProvider: Sendable {
       var id: String { get }            // stable key, e.g. "opencode"
       var displayName: String { get }   // shown in the popover, e.g. "OpenCode"
       func fetchDaily() async throws -> DailyUsage?       // today's total (critical path); nil if no data
       func fetchEnrichment() async -> ProviderEnrichment  // block / week / month (best-effort)
   }
   ```
2. **Register it** in the default `providers:` array — `Sources/PokeTokenBar/Core/UsageStore.swift:319`
   ```swift
   init(providers: [any UsageProvider] = [
       LocalClaudeProvider(), LocalCodexProvider(), LocalGeminiProvider(),
       LocalOpenCodeProvider(), LocalHermesProvider(),
       // + YourAgentProvider(),
   ], ...)
   ```

### Pick the closest reference implementation
- **Log / JSONL source** (append-only session files, like Claude / Codex / Gemini):
  mirror `LocalClaudeProvider` in `Sources/PokeTokenBar/Core/LocalUsageProvider.swift`.
  Reuse the parsing/scan helpers in `LocalUsageReader.swift` — especially
  `period(...)`, `activeBlock(...)`, and `enrichmentScanStart(now:)` for the mtime
  scan window (don't hand-roll the window; that's where a month-boundary bug lived).
- **SQLite / DB source** (like OpenCode / Hermes):
  mirror `LocalOpenCodeProvider` in `Sources/PokeTokenBar/Core/LocalAdditionalUsageProvider.swift`.

## Architecture rules (reviews check for these)

[...]

PR #125, opened 2026-07-26
feat: add Cursor IDE as a usage provider (#115)

Borisserz:
## Summary

Add **Cursor IDE** as a new usage source, implementing the `UsageProvider` extension point from #115.

Cursor stores chat data in a local SQLite database (`~/Library/Application Support/Cursor/User/globalStorage/state.vscdb`). The new `LocalCursorProvider` reads the `cursorDiskKV` table, parsing bubble entries for `tokenCount` (input/output tokens), `createdAt` (ISO 8601 or epoch), and `modelType`.

## Review follow-ups (addressed)

1. **Incremental scan** — `GLOB 'bubbleId:*'` (case-sensitive, index-friendly) + per-database `rowid` high-water mark on the shared cache path (same append-only argument as OpenCode). Refresh scans are `NOT INDEXED` + `rowid > ?`, so SQLite walks only the new rows instead of re-traversing every historical bubble key via the key index (verified with `EXPLAIN QUERY PLAN` in a test).
2. **Watermark safety** — if `MAX(rowid)` drops below the cached watermark (`state.vscdb` recreated / VACUUM / rowid reuse), the provider detects the rewrite and does a full rescan instead of silently skipping usage forever.
3. **End-to-end SQL tests** — temp `CREATE TABLE cursorDiskKV` coverage for table/columns/key filter/watermark/query plan/DB replacement (not only `parseCursorBubble`).
4. **Cost** — Cursor is tokens-only via provider-agnostic `reportsCost` (`false` on `LocalCursorProvider`); per-row cost hidden in the popover, and aggregates (`today`/`week`/`month`, menu bar) fall back to hiding the cost entirely when no enabled provider reports cost — so a Cursor-only setup never shows a misleading `$0.00`.

Nice-to-haves:
- flexible `createdAt` (ISO string **or** epoch seconds/millis)
- `CURSOR_DATA_DIR` override + Cursor Nightly root
- locked static ISO8601 formatters for Cursor cold scans (no per-row formatter allocation)

## Schema (from a real Cursor install)

```sql
CREATE TABLE cursorDiskKV (key TEXT UNIQUE ON CONFLICT REPLACE, value BLOB);
```

Redacted bubble row (other fields omitted):

```json
{
  "type": 2,
  "bubbleId": "<redacted>",
  "createdAt": "2026-01-05T10:09:47.474Z",
  "modelType": null,
  "tokenCount": {
    "inputTokens": 33542,
    "outputTokens": 6856
  }
}
```

Key pattern: `bubbleId:<tabId>:<bubbleId>`

### Core
- **`LocalAdditionalUsageProvider.swift`** — `LocalCursorProvider`, GLOB + rowid-first incremental path with rewrite detection, flexible dates, `CURSOR_DATA_DIR` / Nightly roots, cached ISO8601 formatters
- **`UsageProvider.swift` / `Models.swift` / `UsageStore.swift` / `PopoverView.swift`** — `reportsCost` capability + `costingSnapshots` / `showsCost`

### Tests
- **`CursorUsageTests.swift`** — parser tests + temp-DB e2e (read / date filter / rowid watermark / `EXPLAIN QUERY PLAN` / DB replacement / lowercase-key decoy) + numeric dates + Nightly roots
- **`UsageStoreTests.swift`** — aggregation flow-through + `testCursorContributesZeroToTodayCostTotal` + `testCursorOnlyDoesNotSurfaceZeroDollarCostInMenu`

### Documentation
- **`README.md` / `README.ko.md` / `README.ja.md`** — Cursor in provider lists + Data Sources table (merged with #127 Keychain/outbound-host wording)

## Type of change

- [x] New feature
- [x] Bug fix (review follow-ups on the new provider path)

## UI changes

| Before | After |
| ------ | ----- |
| N/A (new provider) | Cursor tab shows tokens only — no `$0.00` / invented cost row |
| Aggregate cost included mis-priced Cursor models | Cursor excluded from cost aggregates via `reportsCost` |

## Checklist

## Verification

- `swift build` ✅
- Cursor/SQL/cost/parity tests ✅
- CI `build-test` + `secret-scan` ✅
- Merge conflicts with `main` resolved ✅

Closes the Cursor portion of #115.


chattymin:
Three things before merge.

## 1. Full-history scan on every refresh

`SELECT key, value FROM cursorDiskKV WHERE key LIKE 'bubbleId:%'` has no time predicate,
so every row is read and JSON-parsed before its date is known. OpenCode uses
`WHERE time_created >= ?1` (`LocalAdditionalUsageProvider.swift:204`), Hermes
`WHERE started_at >= ?1` (`:235`).

`cursorDiskKV` has no time column, so a direct copy won't work — but the cache-level
incremental path is `.opencode`-only too (`:89`, `:100`), so nothing catches it
downstream. Steady state is O(all history) per cache miss (TTL 30s, `refreshInterval`
120s at `UsageStore.swift:335`), and `state.vscdb` isn't pruned by default.

- `LIKE 'bubbleId:%'` can't use the PK index — default `LIKE` is case-insensitive, so
  SQLite won't rewrite it into a range scan. `GLOB 'bubbleId:*'` will.
- Put `.cursor` on the incremental path. Bubbles are append-only, same argument as the
  comment at `:85-87`. A `rowid` high-water mark or a key-set diff both work.

## 2. The SQL path has no test

All 11 tests hit `parseCursorBubble` (synthetic JSON) plus one missing-path case.
`cursorDatabaseEntries` — table name, column indices, key filter — never runs against a
database. If any identifier is wrong: `prepare_v2` fails → `query` returns `nil` (`:361`)
→ `?? []` → 0 tokens forever, no error surfaced, all 11 tests still green.

The file you're editing does it the other way — `CREATE TABLE` on a temp DB, asserted
end-to-end (`LocalAdditionalUsageTests.swift:22`, `:54`, `:71`, `:107`). Please add the
Cursor equivalent.

I can't verify the schema here (Cursor isn't installed on my machine). Could you paste
`sqlite3 state.vscdb ".schema cursorDiskKV"` and one redacted bubble row into the
description?

## 3. Cost: don't report cost for Cursor at all

Decision on my side: **Cursor reports tokens, not cost.** It's a flat-rate subscription,
so per-token pricing invents money nobody spent — and the family fallbacks in
`ModelPricing.rate(for:)` (`:39-51`) mis-price Cursor's `modelType` anyway
(`claude-3.5-sonnet` → Sonnet 4.6 rates, `gpt-4o` → GPT-5.5 rates, `auto`/`unknown` → $0).

Zeroing the cost isn't enough: `TokenFormatter.cost(0)` renders `"$0.00"`
(`TokenFormatter.swift:27-29`), so the Cursor tab would show a cost *of* $0.00 rather
than no cost. The three aggregate sites are fine with a 0 contribution; it's the
per-provider row (`PopoverView.swift:210`) that needs hiding.

Please do it with a capability flag rather than a `providerID == "cursor"` branch — the
universal paths stay provider-agnostic (#115):

```swift
// UsageProvider.swift — the default keeps all five existing providers unchanged
protocol UsageProvider: Sendable { var reportsCost: Bool { get } }
extension UsageProvider { var reportsCost: Bool { true } }

// LocalCursorProvider
let reportsCost = false

// Models.swift — the default keeps existing call sites unchanged
struct ProviderSnapshot { var reportsCost = true }
```

Then: thread it through the two `ProviderSnapshot(...)` sites (`UsageStore.swift:490`,
`:527`), wrap the cost `Text` at `PopoverView.swift:210` in `if snapshot.reportsCost`,
and skip non-costing snapshots in the four sums (`UsageStore.swift:176-178`, `:188`,
`:190`, `PopoverView.swift:197-199`). A single
`var costingSnapshots: [ProviderSnapshot] { snapshots.filter(\.reportsCost) }` on the
store keeps those four from drifting apart later.

Please add a test that Cursor contributes 0 to `todayCostTotal`.

## Nice to have

- `createdAt` accepts only string ISO-8601. `parseOpenCodeMessage` uses `dateValue(...)`
  (`:340-344`), which takes numbers or strings, seconds or millis.
- `defaultCursorRoots` is hardcoded — no env override (cf. `OPENCODE_DATA_DIR` `:118`,
  `HERMES_HOME` `:147`), no Cursor Nightly.
- `parseISO8601` builds a formatter per row. Less important once (1) lands.

---

Please fill in the PR template checklist on the next push. 
@Borisserz

Borisserz:
Addressed the three merge blockers in `fe4068e`:

1. **Full-history scan** — switched to `GLOB 'bubbleId:*'` and put `.cursor` on the incremental cache path with a **per-database `rowid` high-water mark** (Stable + Nightly don’t share watermarks).
2. **SQL path tests** — added temp-DB end-to-end coverage (`CREATE TABLE cursorDiskKV …`) for read / date filter / watermark. Schema + redacted bubble sample are now in the PR description.
3. **Cost** — `reportsCost` capability flag (default `true`; Cursor `false`), threaded into `ProviderSnapshot`, per-provider cost UI gated in the popover, aggregates via `costingSnapshots`. Added `testCursorContributesZeroToTodayCostTotal`.

Also covered the nice-to-haves: flexible `createdAt`, `CURSOR_DATA_DIR` + Nightly root, reuse of `ISO8601Parser`.

Ready for another look when you have a moment.

Borisserz:
Follow-up after conflict resolution with `main` (#126/#127):

- Merged upstream and resolved `README.md` / `README.ko.md` / `README.ja.md` — Cursor rows kept, Keychain/outbound-host wording from #127 preserved.
- Nice-to-haves tightened: locked static ISO8601 formatters for Cursor cold scans; tests for epoch-seconds `createdAt` and default Stable + Nightly roots (`CURSOR_DATA_DIR` still overrides).

PR checklist + schema sample remain filled in the description. Ready for another look once CI runs.

Borisserz:
Hardening pass after an independent review of the incremental path:

1. **Query plan** — incremental SQL now uses `NOT INDEXED … WHERE rowid > ?1 AND key GLOB 'bubbleId:*'` so refreshes walk new rowids instead of re-scanning the whole `bubbleId:*` key range. Regression: `testCursorIncrementalPrefersRowIDPlan`.
2. **Watermark safety** — if `MAX(rowid) < watermark` (DB rewrite / VACUUM), we full-rescan and discard the stale in-memory set (`didReset`). Regression: `testCursorWatermarkResetsWhenDatabaseIsReplaced`.
3. **`$0.00` leakage** — `showsCost` (`!costingSnapshots.isEmpty`) gates menu cost, popover header cost, and week/month cost labels. Regression: `testCursorOnlyDoesNotSurfaceZeroDollarCostInMenu`.
4. **GLOB filter** — temp-DB test now includes a lowercase `bubbleid:*` decoy that `LIKE` would incorrectly accept.

Ready for re-review.

chattymin:
All three blocking items and the three nice-to-haves are addressed, and the `EXPLAIN QUERY PLAN` regression guard is a better answer than what I asked for — it pins the thing that actually matters (that the incremental path walks rowids and not the key index) instead of pinning the SQL string. Thanks also for the schema and the redacted row.

Confirmed: `GLOB` + per-DB rowid high-water + `NOT INDEXED` + rewrite detection; temp-DB end-to-end coverage; `reportsCost` / `costingSnapshots` / `showsCost` threaded through all four aggregates, the per-provider row, and the menu bar, with the zero-cost test.

I found two defects in the new watermark path. Both reproduce.

## 1. A failed read is treated as a shrink, and the cache is wiped

`LocalAdditionalUsageProvider.swift:374`

```swift
let maxRowID = scalarInt64(database, sql: "SELECT MAX(rowid) FROM cursorDiskKV") ?? 0
let didReset = afterRowID > 0 && maxRowID < afterRowID
```

`query()` returns `[]` when the file is missing but `nil` when open/prepare fails — `?? 0` collapses both into 0, so any read failure satisfies `maxRowID < afterRowID`. `didReset` then makes the cache drop `existing` and keep only this (empty) load, so Cursor's today/week/month all read 0.

Repro: cold-scan a temp DB to establish a watermark, overwrite the file with non-database bytes, query again with that watermark → `didReset=true, entries=0, highWater=0`.

It self-heals on the next cycle (the watermark went to 0, so the next scan is cold), but that's up to one `refreshInterval` — 120s by default — showing zero. `state.vscdb` is being written by Cursor while we read it, so a transient failure isn't hypothetical.

```swift
guard let maxRowID = scalarInt64(database, sql: "SELECT MAX(rowid) FROM cursorDiskKV") else {
    // A failed read is not a shrink — keep the watermark and skip this cycle.
    return CursorDatabaseLoad(entries: [], highWaterRowID: afterRowID, didReset: false)
}
```

A missing file also yields `nil` here, which is fine: those paths never get a non-zero watermark in the first place, so they stay on the cold path exactly as they do today.

Suggested test: `testCursorUnreadableDatabaseKeepsWatermarkAndDoesNotReset` — assert `didReset == false` and that `highWaterRowID` still equals the pre-failure watermark.

## 2. With two roots, a reset in one erases the other's history — and it doesn't come back

`cursorEntries` (`:323-346`) ORs `didReset` across roots, and the cache (`:139-141`) uses that single flag to discard `existing` wholesale. A root that was scanning incrementally only carries rows above its watermark, so its history isn't in the payload that replaces the cache. Its watermark is still high, so nothing re-reads those rows: the undercount persists until a month rollover or an app restart. Unlike (1) this does not self-heal.

Repro: two roots with 5 bubbles each → cold scan (10 entries) → recreate only root B with 1 bubble → re-query with the stored marks → `didReset=true`, payload contains **0** of root A's 5 entries.

The default roots are Cursor and Cursor Nightly, so this needs both installed. (With Nightly absent the path keeps a 0 watermark and stays safe — I checked.)

Since entries don't carry their source path, the simplest correct fix is to make the reset payload complete rather than to make the discard selective:

```swift
var result = scanRoots(sourceRoots, modifiedSince: modifiedSince, marks: afterRowIDByPath ?? [:])
if result.didReset {
    // A partial payload must not replace the cache — rescan every root from scratch.
    result = scanRoots(sourceRoots, modifiedSince: modifiedSince, marks: [:])
}
return result
```

The cache side can stay as it is once the payload is complete.

Suggested test: `testCursorResetInOneRootKeepsOtherRootHistory`.

## Optional, same area

`query()` (`:564`) can't distinguish "finished" from "stopped early" — if `sqlite3_step` returns `SQLITE_BUSY` mid-scan the loop just ends and partial rows are returned as a normal result. The watermark then advances past rows we never read, and those bubbles are lost for the session. Checking for `SQLITE_DONE` after the loop and returning `nil` otherwise closes this alongside (1). Not blocking.

Also worth noting for later: if Cursor ever updates a bubble with `UPDATE ... WHERE key = ?` rather than a plain insert, the rowid wouldn't change and the incremental path would keep the first-seen token count. The `ON CONFLICT REPLACE` in the schema suggests inserts, so I think we're fine, and a cold scan corrects it either way — just something to keep in mind if the numbers ever look low mid-session.
@Borisserz

Borisserz:
Addressed both watermark defects (and the optional incomplete-scan note) in `cbe43da`:

1. **Failed read ≠ shrink** — `MAX(rowid)` now uses `guard let` instead of `?? 0`. Open/prepare failure keeps the prior watermark with `didReset = false`. Covered by `testCursorUnreadableDatabaseKeepsWatermarkAndDoesNotReset`.

2. **Dual-root reset** — when any root reports `didReset`, we cold-rescan *all* roots (`marks: [:]`) and still return `didReset = true` so the provider cache replaces rather than merges a partial payload. Covered by `testCursorResetInOneRootKeepsOtherRootHistory`.

3. **Optional** — `query()` now requires `SQLITE_DONE` after the row loop; otherwise returns `nil` so a mid-scan `SQLITE_BUSY` cannot advance the watermark past unread bubbles.

Local: all `CursorUsageTests` (21) + related UsageStore Cursor cases green.

chattymin:
Merged — thank you for your contribution! 🙇‍♂️

The watermark hardening was thorough: distinguishing a failed read from a shrunken database, and rescanning every root on reset so a partial payload can't replace the cache, are both exactly right, and the regression tests pin them. The `EXPLAIN QUERY PLAN` guard is better than what I asked for — it pins the behaviour that matters rather than the SQL text.

I resolved the README conflicts with `main` (the floating-pet work landed in #124) and added the Cursor row that was missing from the "Works with" table in all three languages.

PR #126, opened 2026-07-27
feat: preselect branching evolution paths

SWARVY:
## Summary

- Preselect and persist a complete evolution path when a companion hatches.
- Render complete linear evolution chains while keeping unrevealed branches hidden as `?`.
- Reveal only the selected branch as the companion evolves.
- Validate and migrate saved companion states so existing progress remains intact.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Branching evolution trees could appear as a misleading linear chain of sibling forms. | Complete linear chains are shown, while unresolved branches are represented by `?`. |
| The next branch was selected only when evolution occurred. | The complete route is selected at hatch time and remains stable across relaunches. |

### Tested cases

| Case | Evolution line |
| --- | --- |
| Linear three-stage evolution<br>Bulbasaur → Ivysaur → Venusaur | Bulbasaur linear evolution |
| Direct branch<br>Wurmple → ? | Wurmple direct branch |
| Multiple direct branches<br>Eevee → ? | Eevee multiple branches |
| Delayed branch<br>Oddish → Gloom → ? | Oddish delayed branch |

## Checklist

### DIFF
--- a/Sources/PokeTokenBar/Core/CompanionStore.swift
+++ b/Sources/PokeTokenBar/Core/CompanionStore.swift
@@ -40,15 +40,20 @@ final class CompanionStore {
     private let clock: () -> Date
     private let fileURL: URL
     private var rng: any RandomNumberGenerator
+    private let dittoDisguiseRollingEnabled: Bool
+    /// 세션 내 활성 개체 교체 감지용. await 뒤 이전 개체의 결과가 새 개체를 덮지 않게 한다.
+    private var activeGeneration = 0
 
     init(provider: any PokeProviding = PokeAPIClient.shared,
          clock: @escaping () -> Date = Date.init,
          fileURL: URL? = nil,
-         rng: any RandomNumberGenerator = SystemRandomNumberGenerator()) {
+         rng: any RandomNumberGenerator = SystemRandomNumberGenerator(),
+         dittoDisguiseRollingEnabled: Bool = AppEnv.isBundledApp) {
         self.provider = provider
         self.clock = clock
         self.fileURL = fileURL ?? Self.defaultURL()
         self.rng = rng
+        self.dittoDisguiseRollingEnabled = dittoDisguiseRollingEnabled
         load()
         if state.active != nil { displayState = .idle }
     }
@@ -116,15 +121,29 @@ final class CompanionStore {
     }
     var tokensToNext: Int { guard let a = state.active else { return 0 }; return max(0, threshold - a.usedAtStage) }
 
-    /// 진화 라인 표시용: 현재까지 경로 + 다음 후보. (id, kind) kind: done/cur/future
-    var lineNodes: [(id: Int, kind: String)] {
+    /// 진화 라인 표시용: 실현된 경로 + 다음 단계 미리보기.
+    /// 유일하게 이어지는 단계 뒤에 분기가 있으면, 그 확정 접두어와 하나의 미지 항목을 함께 보여 준다.
+    /// 분기 후보는 부화 시 계획됐더라도 실제 진화 전까지 하나의 미지 항목으로 숨긴다.
+    var lineNodes: [EvoLineItem] {
         guard let a = state.active, let line = currentLine else { return [] }
-        var out: [(Int, String)] = []
+        var out: [EvoLineItem] = []
         for (i, id) in a.pathIDs.enumerated() {
-            out.append((id, i < a.stageIndex ? "done" : (i == a.stageIndex ? "cur" : "future")))
+            out.append(EvoLineItem(.species(id), i == a.pathIDs.count - 1 ? .current : .done))
         }
-        if let cur = line.tree.node(withID: a.currentID) {
-            for ch in cur.children { out.append((ch.speciesID, "future")) }
+        if let current = line.tree.node(withID: a.currentID) {
+            var node = current
+            var guaranteedPrefix: [EvoNode] = []
+            while node.children.count == 1, let child = node.children.first {
+                guaranteedPrefix.append(child)
+                node = child
+            }
+
+            if node.children.count > 1 {
+                out += guaranteedPrefix.map { EvoLineItem(.species($0.speciesID), .future) }
+                out.append(EvoLineItem(.mystery, .future))
+            } else {
+                out += guaranteedPrefix.map { EvoLineItem(.species($0.speciesID), .future) }
+            }
         }
         return out
     }
@@ -280,7 +299,16 @@ final class CompanionStore {
             if node.children.isEmpty {
                 graduate(); break
             } else {
-                let next = pickNextChild(node, baseID: a.baseID)
+                let nextIndex = a.stageIndex + 1
+                guard a.plannedPathIDs.indices.contains(nextIndex) else {
+                    AppLog.write("evolve: invalid planned path index \(nextIndex) for base \(a.baseID)")
+                    break
+                }
+                let nextID = a.plannedPathIDs[nextIndex]
+                guard let next = node.children.first(where: { $0.speciesID == nextID }) else {
+                    AppLog.write("evolve: planned child \(nextID) is invalid for current \(a.currentID)")
+                    break
+                }
                 state.active!.pathIDs = Array(a.pathIDs.prefix(a.stageIndex + 1)) + [next.speciesID]
                 state.active!.stageIndex += 1
                 state.active!.usedAtStage = a.usedAtStage - thr   // 초과분 이월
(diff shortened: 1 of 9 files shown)


chattymin (commented):
Reviewed the full diff, built and ran the branch locally (`swift test`: 339 tests, 0 failures), and verified the balance-sensitive parts against live PokeAPI data.

Two things I want to call out as verified-good, because they are the parts that would have been easy to get wrong:

- **Preselecting at hatch is behaviourally equivalent to picking at evolve time.** `collectedFinals` is only mutated in `graduate()` (`Core/CompanionStore.swift:384`), which also clears `state.active` — so the set cannot change during a single companion's lifetime, and the "prefer uncollected finals" heuristic sees identical input either way. No dex-completion regression, only a change in RNG draw order.
- **`totalForms` moving from `tree.depth` to `plan.count` does not shift any real user's thresholds.** I pulled every evolution chain from the PokeAPI GraphQL endpoint: within the animated-sprite range (<=649) there are 11 branching lines (43, 60, 79, 133, 236, 265, 280, 290, 361, 366, 412) and **zero** with asymmetric branch depth, so `plan.count == tree.depth` always holds in production. The asymmetric handling is a latent correctness improvement (on a synthetic asymmetric line the graduation total goes 375M -> 750M, i.e. back to the intended value) rather than a live balance change.

Also worth noting that the `activeGeneration` / `latest` re-read in `loadCurrentLine` is not incidental hardening — it is required by the refactor. The old code assigned individual fields, whereas `state.active = normalizedEvolutionState(...)` writes the whole struct, so reusing the pre-`await` snapshot would drop concurrent `usedAtStage` and `nature` changes. `testLineLoadPreservesUpdatesMadeWhileProviderIsSuspended` reproduces exactly that trigger.

The inline comments below are all minor — none of them block merge.

Sources/PokeTokenBar/Core/CompanionStore.swift
```diff
@@ -116,15 +121,29 @@ final class CompanionStore {
...
+    /// 진화 라인 표시용: 실현된 경로 + 다음 단계 미리보기.
+    /// 유일하게 이어지는 단계 뒤에 분기가 있으면, 그 확정 접두어와 하나의 미지 항목을 함께 보여 준다.
+    /// 분기 후보는 부화 시 계획됐더라도 실제 진화 전까지 하나의 미지 항목으로 숨긴다.
+    var lineNodes: [EvoLineItem] {
         guard let a = state.active, let line = currentLine else { return [] }
-        var out: [(Int, String)] = []
+        var out: [EvoLineItem] = []
         for (i, id) in a.pathIDs.enumerated() {
-            out.append((id, i < a.stageIndex ? "done" : (i == a.stageIndex ? "cur" : "future")))
+            out.append(EvoLineItem(.species(id), i == a.pathIDs.count - 1 ? .current : .done))
```
chattymin: `lineNodes` decides "current" from `pathIDs.count - 1`, while `currentID` (`Core/CompanionModel.swift:338`) decides the same thing from `min(stageIndex, pathIDs.count - 1)`. Two different rules for one concept.

This is safe today: every writer — `hatch`, the evolve branch below, `revealDitto`, `normalizedEvolutionState` — keeps `stageIndex == pathIDs.count - 1`, and `lineNodes` returns early while `currentLine == nil`, so it only ever runs after normalization. But `currentID` explicitly clamps against that invariant breaking and this line does not, so any state that violated it would put the current-marker dot on a different species than `currentID` reports.

`i == a.stageIndex` would make both derive from the same field.

Sources/PokeTokenBar/Core/CompanionStore.swift
```diff
@@ -280,7 +299,16 @@ final class CompanionStore {
...
+                let nextIndex = a.stageIndex + 1
+                guard a.plannedPathIDs.indices.contains(nextIndex) else {
+                    AppLog.write("evolve: invalid planned path index \(nextIndex) for base \(a.baseID)")
+                    break
+                }
+                let nextID = a.plannedPathIDs[nextIndex]
+                guard let next = node.children.first(where: { $0.speciesID == nextID }) else {
+                    AppLog.write("evolve: planned child \(nextID) is invalid for current \(a.currentID)")
+                    break
+                }
```
chattymin, line 318: These two guards turn an invalid plan into a silent stall rather than a visible failure: `usedAtStage` stays above the threshold, `progress` pins at 1.0 and `tokensToNext` at 0, and the companion never evolves again. Only `AppLog` records it — there is no user-visible signal, and the popover keeps showing a full progress bar.

Unreachable today, since `normalizedEvolutionState` rebuilds a valid plan on every line load, so this is pure defense. Worth noting though that the pre-PR code could not stall at all: `pickNextChild` always returned some child.

Falling back to `pickPlannedChild(node, baseID: a.baseID)` (and persisting the repaired plan) instead of `break` would remove the stall state entirely and match the previous behaviour.

Sources/PokeTokenBar/UI/CompanionView.swift
```diff
@@ -141,7 +141,8 @@ struct SpriteView: View {
 /// 팝오버 전체가 좌우로 잘린다**(진화줄뿐 아니라 탭바·합계까지). `maxWidth` 를 주면 그 폭 안에서
 /// 가로 스크롤한다 — 썸네일 크기는 유지하고, 가장자리 페이드 + 셰브론으로 스크롤 가능함을 알린다.
 struct EvoLineView: View {
-    let nodes: [(id: Int, kind: String)]
+    let nodes: [EvoLineItem]
+    let language: AppLanguage
```
chattymin: `language` is threaded through every call site for a single accessibility label (line 319). Two narrower options: take the already-resolved string (`mysteryLabel: String`), or read the language from the environment.

Not blocking — just noting the parameter's reach is much wider than its use.

Tests/PokeTokenBarTests/EvoLineLayoutTests.swift
```diff
@@ -73,8 +73,8 @@ final class EvoLineLayoutTests: XCTestCase {
     /// rowWidth 는 레이아웃과 같은 식이어야 한다(측정값과 일치 — 어긋나면 스크롤 판정이 틀어진다).
     func testRowWidthMatchesRenderedWidth() {
         for count in 1...9 {
-            let nodes = (1...count).map { (id: $0, kind: "done") }
-            let rendered = renderedWidth(EvoLineView(nodes: nodes), proposing: 4000)
+            let nodes = (1...count).map { EvoLineItem(.species($0), .done) }
+            let rendered = renderedWidth(EvoLineView(nodes: nodes, language: .en), proposing: 4000)
```
chattymin: This file converted every case to `.species`, so `case .mystery` — the render branch this PR adds to `EvoLineView` — has no width coverage. `rowWidth`/`needsScroll` assume every cell is `thumb` wide, and the mystery cell only gets that from an explicit `.frame(width: thumb, height: thumb)` that nothing locks.

I measured it on this branch and it does match today (1...6 cells rendering 40/94/148/202/256/310pt, identical to `rowWidth`), so this is a coverage gap rather than a defect. Extending this loop covers it:

```swift
let nodes = (1...count).map { i in
    i == count ? EvoLineItem(.mystery, .future) : EvoLineItem(.species(i), .done)
}
```

Related, and currently unreachable: when `names != nil` a mystery cell has no name label, so it renders 6pt narrower than `rowWidth(hasNames: true)` assumes (measured 136 vs 142). `names` is only passed from `DexEntryRow`, which never contains mystery nodes — but it does mean the width formula is not content-agnostic.

Sources/PokeTokenBar/Core/CompanionStore.swift
SWARVY: Addressed in `2b40ce0`. `lineNodes` now builds realized items through `realizedLineItems(pathIDs:stageIndex:)`, so the current marker derives from `stageIndex`, matching `currentID`. Added `testRealizedLineItemsUsesStageIndexForCurrentMarker` for the divergent-invariant case. Thanks for catching the split source of truth.

Sources/PokeTokenBar/Core/CompanionStore.swift
SWARVY: Addressed in `2b40ce0`. The evolution path now falls back to `pickPlannedChild` when the stored index or child is invalid, rebuilds the remaining route, and persists the repaired `plannedPathIDs` and `totalForms` before continuing. Added focused repair coverage; the existing reload-invalid-plan test also continues to verify normalization. This removes the full-progress stall state.

Sources/PokeTokenBar/UI/CompanionView.swift
SWARVY: Addressed in `2b40ce0`. `EvoLineView` now accepts `mysteryLabel: String`, and each caller resolves `store.l.unknownNextEvolution`, keeping localization outside the view API.

Tests/PokeTokenBarTests/EvoLineLayoutTests.swift
SWARVY: Addressed in `2b40ce0`. The rendered-width matrix now places a `.mystery` item in every count, and a new named-row test verifies that a mystery cell reserves the same column width. The row now fixes every cell to `thumb + nameSlack` whenever names are present. Full verification passes: `swift build` and `swift test` completed with 342 tests executed, 5 skipped, and 0 failures.

chattymin:
Follow-up on the `stageIndex` / `currentID` thread — optional, not reachable today.

  `lineNodes` still derives "current" two ways: the marker from `stageIndex` (line 129, unclamped), and the
  preview's starting node from `currentID` (line 130), which clamps with `min(stageIndex, pathIDs.count -
  1)`. With `pathIDs: [1,2], stageIndex: 5` the marker disappears while the preview still resolves to `2` —
  the mirror of the case this commit fixed. `repairedPlan` in the same commit does clamp, so the two new
  helpers take opposite stances on the same input.

  Rather than repeating the clamp, it could be closed at the decode boundary. `MonState.init(from:)`
  validates `pathIDs` (`Core/CompanionModel.swift:366`) but reads `stageIndex` raw at line 374:

  ```swift
  let decoded = try c.decode(Int.self, forKey: .stageIndex)
  stageIndex = min(max(0, decoded), pathIDs.count - 1)
  ```

  Clamp rather than throw: an out-of-range stageIndex still has a valid realized path behind it. No
  behaviour change for valid saves, and realizedLineItems can then stay exactly as written.

  Fine as a follow-up issue if you'd rather keep this PR scoped.
  @SWARVY

SWARVY:
Addressed in `0009956`. `MonState` now clamps decoded `stageIndex` at the persistence boundary to `0...(pathIDs.count - 1)`, preserving a valid realized path without changing valid saves. Added regression coverage for both upper and lower out-of-range values. Verified with `swift build` and `swift test` (343 executed, 5 skipped, 0 failures).

chattymin:

Issue #128, opened 2026-07-28
Codex: forked/subagent rollouts replay the parent's token_count history, inflating daily usage several-fold (observed 5.8–7.2×)


dkswhale:
## Summary

When Codex CLI spawns **subagent / forked sessions**, each child rollout file **replays the parent's entire conversation history — including every `token_count` event — re-stamped at fork time**. PokeTokenBar sums all `token_count` events without any cross-file dedup, so on days with heavy subagent usage the Codex numbers inflate several-fold (I measured **5.8×–7.2×**), and yesterday's turns bleed into today's bucket. My menu bar showed **12.6B cache-read tokens for a single day**; the real figure was ~1.75B.

## Environment

- Codex CLI `0.145.0` (`payload.cli_version` in rollout files), heavy subagent/orchestration usage
- Parsing logic confirmed against current `main` — `LocalUsageReader.parseCodexFile` / `parseCodexLine`

## Root cause

Codex fork/subagent rollouts are identifiable in the first line (`session_meta`):

```json
{ "type": "session_meta", "payload": {
    "forked_from_id": "019fa149-...", "parent_thread_id": "019fa149-...",
    "thread_source": "subagent", "history_mode": "legacy", "cli_version": "0.145.0" } }
```

The child file contains a full copy of the parent's history. All replayed `token_count` events carry **new timestamps (fork time)**, not their original ones.

`parseCodexFile` creates one entry per `token_count` line with id `codex|<basename>|<turn>`, so:

1. **Multiplication** — a parent with N turns forked K times is counted K+1 times. Orchestrators that fork repeatedly compound this throughout the day.
2. **Day shifting** — because replays are re-stamped, turns that actually ran yesterday land in today's `localDay` bucket.

## Measured impact (2026-07-28, KST, my machine)

| | files (mtime today) | fork files | token_count events | replay events | naive cacheRead | replay-removed |
|---|---|---|---|---|---|---|
| 00:24 | 34 | 32 | 46,062 | 35,150 | 4.89B | 0.84B (**5.8×**) |
| 16:36 | 93 | 87 | 102,548 | 87,656 | 12.60B | 1.75B (**7.2×**) |

The ratio grows during the day as forks accumulate.

## How replays can be told apart

[...]

PR #129, opened 2026-07-29
fix: exclude replayed usage from forked Codex sessions

SWARVY:
refs #128 

## Summary

- Detect forked Codex rollout sessions and exclude the leading rapid `token_count` replay burst from the parent conversation.
- Keep the first real child turn and all subsequent usage entries.
- Invalidate only the persisted Codex usage cache after this parser change, so existing stale replayed entries are re-parsed.
- Add regression coverage for direct parsing, cached parsing, and cache migration.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

The popover layout is unchanged; the displayed Codex usage now excludes replayed parent-session history.

| Before | After |
| ------ | ----- |
| image | image |

## Checklist


chattymin:
Two failure modes in the replay detection. Both reproduced.

## 1. Blocker — the trim does nothing if the burst starts late

`replayTimestamp` is seeded from the `session_meta` timestamp, and the first `token_count` that is ≥2s away sets it to `nil` for the rest of the file. If the burst itself starts more than 2s after `session_meta`, that first event *is* a replay and the check concludes the opposite.

Fork meta at `01:00:00.000`, first replay at `01:00:03.000`, seven more at 10ms intervals, one real turn 40s later:

```
### A: kept=9  outputs=[1, 1, 1, 1, 1, 1, 1, 1, 99]
     expected: kept=1
```

Not one replay dropped. Disk latency or a large parent history being serialized is enough to push that first write past 2s, and there's no signal when it happens.

## 2. Real turns are swallowed

Trimming stops at the first gap ≥2s, not at the end of the burst, so real turns spaced under 2s apart are dropped in a chain.

Three replays at 10ms, then three real turns 1.5s apart, then one 56s later:

```
### B: kept=1  outputs=[44]
     expected: [11, 22, 33, 44]
```

(My "worst case is one lost turn" in #128 was wrong — the loss runs until the first ≥2s gap. You built to that spec.)

In your own fixture the last two child events are **2.07s** apart — 0.07s of margin.

## Root cause

The 2s threshold answers two questions on different scales: "still inside the burst?" (tens of ms) and "has the burst ended?" (seconds). That's why it fails open one way and closed the other.

## Fix

**Anchor on the first `token_count` instead of `session_meta`.** In a forked rollout the first `token_count` is a replay by construction, so distance from `session_meta` stops mattering. Cost: a fork whose parent had zero turns loses one real event — bounded at one.

**Threshold 2s → ~1s.** Burst-internal gaps are tens of ms; the smallest burst-to-real-turn gap measured in #128 was 7.85s. 1s sits 20–30× above the former and 8× below the latter.

## Tests

Every case in the PR uses the same shape — 10ms burst, ≥3s gap — so none of them exercise either failure mode. Please pin both:

1. first replay ≥2s after `session_meta`, rest of the burst at 10ms → only the real turn survives
2. burst, then real turns 1.5s apart → all real turns survive

## Minor

- The fixture is a manual fork (`thread_source: "user"`, `0.142.5`); the issue is about `thread_source: "subagent"` on `0.145.0`. Detection covers both, but the reported shape isn't pinned. The reporter offered sanitized samples in #128.
- Detection only inspects line 1 — if Codex ever emits a record before `session_meta`, it turns off silently.

@SWARVY

SWARVY:
Implemented in b38bb94.

- The first token_count in a detected fork now anchors replay trimming, so session metadata write timing cannot bypass it.
- Replay mode uses a 1-second inter-event threshold and exits permanently at the first longer gap, preserving subsequent rapid real turns.
- Fork metadata is latched while scanning leading non-token records, covering an initial preamble without letting replayed parent metadata undo detection.
- Bumped the Codex parser cache revision from 1 to 2 so entries cached under the previous rule are re-parsed.
- Added regressions for a delayed replay burst, 1.5-second-spaced real turns, the 0.145.0 subagent parent_thread_id shape, and the prior cache revision.

The no-parent-token fork trade-off remains bounded to the first child token_count, as noted.

swift test: 351 tests run, 5 skipped, 0 failures.

chattymin:

Issue #132, opened 2026-07-29
Codex: derive entry ids from content so forked/subagent replays collapse without a time heuristic


chattymin:
## Summary

Codex entry ids are positional — `codex|<basename>|<turn>` (`LocalUsageReader.swift:179`). Two files holding the same event therefore produce two different ids, which is why a forked or subagent rollout replaying its parent's history was counted twice (#128).

#129 fixed the symptom by trimming the leading replay burst: in a file whose `session_meta` carries `forked_from_id` / `parent_thread_id`, events are dropped until the first gap of a second or more. That works, but it is a timing guess, and "how many seconds means replay" has no correct answer. Deriving the id from the event's content removes the question instead of tuning it.

Claude Code has never had this problem for the same reason — it dedups on `(message.id, requestId)` with no file component (`LocalUsageReader.swift:7`), so a message copied into a second file collapses onto the same entry.

## Why the current heuristic keeps moving

Both failure modes below were reproduced during review of #129 and are now guarded by tests, but they are the shape of the problem rather than one-off bugs:

- **Fails open.** The trim was anchored on the `session_meta` timestamp. If the burst started more than the threshold after it (disk latency, a large parent history being serialized), the first replay looked like a real turn and *nothing* was dropped — 9 events kept where 1 was expected.
- **Fails closed.** Trimming stopped at the first gap over the threshold rather than at the end of the burst, so real turns spaced under the threshold apart were swallowed in a chain — 1 event kept where 4 were expected.

The threshold is now 1s anchored on the first `token_count`, which clears the measured data in #128 (burst-internal gaps in the tens of ms, smallest burst-to-real-turn gap 7.85s). It is still a value chosen against one sample.

## What the data actually gives us

A `token_count` line carries no session identifier:

```
top:     [timestamp, type, payload]
payload: [type, info]
info:    [total_token_usage, last_token_usage, model_context_window]
```

But the fixture committed in #129 shows the parent's `session_meta` is re-inserted into the child file ahead of the replayed events:

```
1   session_meta   { id: child,  forked_from_id: parent }
2   session_meta   { id: parent }          <- parent metadata, re-inserted
3+  token_count x8                          <- parent history, replayed
11  token_count                             <- the child's own first turn
```

[...]

PR #137, opened 2026-07-30
fix: identify Codex turns by content instead of file position

LeeTaek:
Summary

Codex usage entries were identified by file and line position. This caused:

- duplicated usage when the same state was re-recorded or copied into a fork;
- missing usage when the #129 timing heuristic removed the first real subagent turn.

This PR identifies turns from their session and full usage state, and resolves fork replay by
comparing the child with its parent rollout.

Refs #132. Replaces the #129 timing heuristic as the primary resolution path for #128.

## Why

Real subagent captures from Codex CLI `0.142.5` and `0.145.0` have a different shape from a
manual `codex fork`:

| | Manual fork | Subagent |
| --- | --- | --- |
| Parent `session_meta` inserted | Yes | Yes |
| Parent `token_count` replayed | Yes | **No** |
| Events after the parent metadata | Replayed parent history | Child's own turns |

The old heuristic treated both shapes as replay and dropped the subagent's first turn:

| Fixture | Counted on `main` | Counted after |
| --- | ---: | ---: |
| Subagent `0.142.5` | 23,291 | 46,353 |
| Subagent `0.145.0` | 21,722 | 43,180 |

The committed fixtures use fixed synthetic IDs and contain no prompts, responses, paths, Git
metadata, or instructions.

## What changed

- Parse `session_meta.payload.id` before `session_id`; subagent `session_id` points to the parent.
- Compare a child's leading full usage-state sequence with its parent's resolved history. Only an
  exact matching prefix is replay; a non-matching prefix keeps all child events.
- Resolve parents outside the query's mtime window as dependencies, including fork-of-fork chains.
- Generate canonical IDs from:

  ```text
  codex|<session id>|<epoch>|<cumulative vector>|<last vector>
  ```

- Deduplicate canonical IDs after ownership resolution, keeping the earliest timestamp.
- Drop consecutive re-records only when both the cumulative and last-usage vectors are identical
  in the same session.
- Cache parsed rollouts instead of final entries and bump `codexParserVersion` from 2 to 4.

The full vectors avoid collisions when scalar totals match but token categories differ. The last
vector also preserves the zero-delta turn in `Fixtures/CodexFork/child.jsonl`.

## Fallback and compatibility

- If a manual fork's parent is missing or full usage state is unavailable, the existing one-second
  trim remains as a recovery fallback.
- Subagents are exempt from the timing fallback because both captured versions contain no replay.
- Legacy records without cumulative usage or a session ID keep their positional IDs.
- Gemini and UI code are unchanged.

## Tests

- `swift build`
- `swift test` — 434 tests executed, 5 skipped, 0 failures

Coverage includes both subagent versions, manual and sibling forks, fork-of-fork, zero-delta turns,
cumulative resets, non-fork parity, canonical cross-file deduplication, fallback behavior, and
parsed-rollout cache migration.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## Checklist

- [x] No UI changes


chattymin:
Thanks for the detailed writeup — the shape comparison table and the per-fixture counts made this a lot easier to review. @LeeTaek 

  I checked the branch out and ran it against my own `~/.codex/sessions` (57 rollouts, 6 with a parent link, 4 subagents) rather than relying on the fixtures
  alone. The core fix holds up, and the reconciliation is exact:

  | | entries | tokens |
  | --- | ---: | ---: |
  | raw `token_count` events | 533 | 53,868,709 |
  | `main` | 528 | 53,773,760 |
  | this branch | 522 | 52,923,086 |

  Counting — with a separate script, not this code — the events whose cumulative *and* last usage both repeat the previous event in the same session gives 10
  events / 945,623 tokens. `53,868,709 − 945,623 = 52,923,086`, which matches this branch exactly, so the only usage dropped here is genuinely re-recorded
  state. In the other direction, 4 of the 5 events `main` was dropping were real subagent first turns (94,949 tokens). Both halves of the problem reproduce.

  A few other things confirmed on real data: subagent `session_meta` does carry `id` = child with `session_id` = parent, the child meta is always the first
  line, `swift test` gives 435 tests / 5 skipped / 0 failures, the cache path and the reader path produce identical output, and day totals are stable across
  query window sizes.

  I also like that the cache now stores parsed rollouts instead of final entries. The old blob froze a cross-file verdict behind a `(path, mtime, size)` key,
  which meant the verdict could never be revisited once the neighbouring evidence changed. Re-evaluating resolution on every refresh is the right boundary.

  One issue I'd like to see addressed before merge, plus three smaller notes.

  ## Orphaned parent triggers a full-tree scan on every refresh

  `LocalUsageCache.swift:233-236` — when `indexedMatches` is empty, the fallback filter calls `codexRolloutSessionID(at:)` on every unresolved file, i.e.
  opens and reads 64 KB from every rollout outside the mtime window. `searchedParentIDs` (line 223) only dedupes within a single call, and nothing records
  "this parent id has no file", so the scan repeats on every refresh tick.

  Measured on my machine (57 rollouts):

  | query window | `main` | this branch |
  | --- | ---: | ---: |
  | contains an orphaned fork | 1.0 ms | **129.7 ms** |
  | no fork (control) | 7.2 ms | 7.8 ms |

  The trigger is ordinary — the parent rollout only has to have been deleted or archived. It already fires on my machine: `019ecd9d` has `forked_from_id:
  7a7656f1` and no surviving file carries that session id. Two things make it worth fixing rather than accepting: the cost scales with the *total* rollout
  count rather than the window size, and it runs inside the `LocalUsageCache` actor, so it serializes every usage refresh behind it. At a few thousand
  rollouts this is seconds per tick.

  Caching the negative result should be enough — persisting the set of parent ids that could not be resolved, or keeping a `path → sessionID` index in the
  snapshot, would bound the content scan to at most once per parent id. The same loop exists in `LocalUsageReader.swift:314-328` as well, so both copies need
  it.

  ## Smaller notes

  - **`LocalUsageReader.swift:352-356`** — `read(upToCount: 64 * 1024)` followed by `String(data:encoding:.utf8)` returns `nil` if the 64 KB cut lands
  mid-codepoint, which is reachable with a non-ASCII `cwd`. The probe then reports "no session id" and resolution silently degrades to the timing fallback.
  The same happens if the first line exceeds 64 KB. Trimming to the last newline before decoding, or reading line-wise, would avoid it.

  - **Duplicated resolution loop.** `LocalUsageReader.codexEntries` (298-330) and `LocalUsageCache.collectCodexRollouts` (172-246) implement the same closure
  expansion over two different file representations. Production only reaches the cache path (`LocalUsageProvider` → `LocalUsageCache.shared.codexEntries`), so
  several of the new reader tests exercise a sibling implementation rather than the shipping one. They agree today — I verified that on real data — but a fix
  applied to one copy would leave the other green and the user unaffected. Extracting the expansion behind a small file-source abstraction would collapse
  them.

LeeTaek:
Thanks for the review — checking the branch against your own `~/.codex/sessions` rather than the fixtures made it clear what was left to fix. I've split the response into three commits:

| commit | |
| --- | --- |
| `5e8784c` | `fix: probe Codex session metadata line-wise` |
| `52f8992` | `refactor: share Codex parent rollout discovery` |
| `d6267ca` | `perf: index Codex rollout session ids so orphaned parents don't rescan` |

## 1. Repeated full-tree scan for an orphaned parent

Of the two options you suggested I went with the `path → sessionID` index. Persisting a negative set of parent ids risks never finding a parent that is later restored or synced back, whereas a per-file index is invalidated naturally by `(path, mtime, size)`.

Lookup order:

```text
valid blob → session id index → filename hint → content probe
```

`sessionID == nil` is stored as a completed probe result too, so a lookup answers with one of three states:

- a session id
- probed, and the file has none
- not probed yet

That way a file without `session_meta` is not reopened on every refresh. A full parse also records into the index, and the index is kept for as long as the file exists — parsed blobs are pruned after 40 days, and finding parents older than that is the whole point of the index.

The performance contract is probe count rather than elapsed time. Measured on a temporary copy of the real tree with an orphaned fork planted (109 rollouts):

```text
cold: 108 probes (~0.5s)
warm:   0 additional probes (~4ms)
```

The warm path used to repeat the full scan. Both the daily and the enrichment query in the production pipeline now reuse the same persisted index.

## 2. Fixed 64 KiB UTF-8 decode

The old probe decoded the whole 64 KiB prefix at once, so a boundary landing mid-codepoint made the entire probe return `nil`. It is now a line-wise scanner with a read budget:

- handles a metadata line larger than 64 KiB
- preserves UTF-8 code points split across a read chunk boundary
- keeps the existing stop conditions
- caps the whole scan at 1 MiB

A line that is not valid UTF-8 stops the search rather than being skipped. When the full parser fails to decode a file, a probe that skipped ahead could read the re-recorded metadata that follows, and the two paths would then disagree on the session id.

I did not observe a first line over 64 KiB in my own tree, but it is a realistic risk: the largest `session_meta` line here is ~46 KB, most of it `dynamic_tools` (grows with the number of MCP tools) and `base_instructions` (grows with user configuration).

In an opt-in test over the real tree, the probe and the full parser return the same session id for all 109 rollouts. The old fixed-prefix decode returned `nil` for 14 of them.

## 3. Duplicated resolution loop

Both the reader and the cache path now go through `LocalUsageReader.expandCodexParentClosure`, sharing the file representation `CodexRolloutFile { url, mtime, size }` and injecting only three things:

- how a file is parsed
- which session ids are already known
- the content probe

The production cache path uses the persisted index. The direct reader path has no persistence layer, but it runs the same resolution algorithm, so a later change to the resolver cannot apply differently to the two paths.

To confirm the refactor changed no behaviour, I checked the parent commit out into a separate worktree and ran both against the same Codex tree: the reader aggregates and the Codex snapshot through the full pipeline were identical.

## Related fixes made along the way

- A filename hint that fails payload verification no longer suppresses the content probe.
- A cached blob's session id is only trusted while the file's mtime and size still match.

## Tests

```text
swift test: 451 executed, 7 skipped, 0 failures
```

New tests: 7 for the line-wise probe, 8 for the cache index, 2 opt-in parity. For each defect I confirmed that the core regression test fails when the corresponding fix is reverted. These three commits do not raise the existing `codexParserVersion = 4`, so they introduce no additional parsed-blob invalidation.

When you have time, it would be worth re-running your reconciliation. On trees where the old probe failed to find a parent that does exist, the child is now trimmed by exact prefix comparison instead of the timing fallback, so totals can shift. The orphan you mentioned (`019ecd9d → 7a7656f1`) has no parent file at all, so I'd expect no effect there.

chattymin:
Thanks for the quick turnaround — the line-wise probe, the session-id index, and the shared parent discovery all address what I raised. @LeeTaek

I re-checked the branch against paths my own `~/.codex/sessions` doesn't cover (all my rollouts share one CLI version). Two come out worse than `main`, so I'd like those before merge. The rest is pre-existing or an improvement, and I'll file it separately.

**1. A found-but-mismatched parent skips trimming.** `comparableUsagePrefixCount` (`LocalUsageReader.swift:588-590`) returns `0`, not `nil`, when the first vectors differ, so `if let bestParentMatch` (`:529`) succeeds with `replayCount: 0` and nothing is trimmed. The timing fallback (`:537`) only runs when no parent was found — finding the parent ends up worse than not finding it.

Not hypothetical: your own fixtures differ here — `parent-v145.jsonl` carries `cache_write_input_tokens`, `parent.jsonl` doesn't. A parent recorded before a CLI upgrade and a fork taken after it disagree at index 0, so the whole replayed prefix counts as the child's turns (repro: `[110, 220, 220, 1098]`, expected `[220, 1098]`). `main` trims these via the 1s heuristic, which makes it a regression rather than a leftover gap.

Falling through to the fallback when `replayCount == 0 && rollout.parentSessionID != nil` should cover it; subagents are unaffected since `fallbackReplayCount` already returns `0`. The regression test wants the **parent-present, vectors-disagree** case — the suite doesn't have it.

**2. Cumulative usage can trap.** `CodexUsageVector` reads its fields through `intValue` (`:144-149`), an unchecked `Int(d)` at `:1022`. A line with `"input_tokens": 1e30` kills the process; the same line on `main` parses fine, since `main` never read `total_token_usage`. The file stays on disk, so it recurs on every relaunch until deleted by hand. `intOrNil` (`:1037`) already clamps for this and drops in. Happy to push it myself if that saves a round.

Two smaller notes:

- The probe's negative result is cached without distinguishing why it's negative. `try?` at `:445,452` swallows read errors and the `nil` is persisted (`LocalUsageCache.swift:260-262`), so one transient failure pins a file as "no session id" until its `(mtime, size)` changes. Remembering only "no metadata", not "couldn't read", would fix it — note `testCodexNegativeSessionIDProbePersistsAcrossRefreshesAndInstances` locks in the current behaviour.
- `refs #132` rather than `Closes`. Attributing to the child instead of the parent is a fair call given what you confirmed about subagents, but the issue also asked to drop the timing heuristic. I'll re-scope it.

Everything else — the per-line `JSONSerialization` cost, the parent-hint substring match, id collisions when a session id reappears in a second file — is no worse than `main` or not reachable today, and some predates this PR. Those and the other unclamped `intValue` calls are on me, not you.

LeeTaek:
Thanks for the second pass — both regressions and both notes are addressed, in three commits:

* `586c765`: when the parent is found but the comparable usage prefix is 0, a manual fork now falls back to the timing trim. Subagents keep all of their real turns as before.
* `78ad95a`: the cumulative usage vector this PR introduced is read through `intOrNil`, so a value like `1e30` no longer traps. The pre-existing `last_token_usage` call sites are left out of scope.
* `26fde6c`: a failed probe read is no longer written to the index and is retried on the next refresh; within one collection it is not retried. Only a genuine "no metadata" is stored as a negative, and the session-id index moved to v2 to drop the v1 entries that cannot be told apart. Parsed rollout blobs are kept.

Regression tests plus snapshot and index-migration tests were added.

```
swift build: passed
swift test: 456 executed / 7 skipped / 0 failures
real rollouts: 109, probe/parser mismatches: 0
orphan scan: cold 107 probes → warm +0 (~5ms)
```

Running `d6267ca` and the current HEAD back to back against the same `~/.codex/sessions` gives an identical result over every entry:

```
entries: 3,892 → 3,892
tokens: 446,299,747 → 446,299,747
```

One correction: in the fixtures `cache_write_input_tokens` is absent on the older version and `0` on the newer one, so the presence of the field alone does not make the vectors differ. The actual trigger is a different value being recorded in that category.

The PR body now says `Refs #132`. If you get a chance, a re-run of your reconciliation against the current HEAD would be welcome.

chattymin:

PR #143, opened 2026-08-05
fix: keep companion progress moving after usage snapshot drops

ite476:
## Summary

- Rebase the companion's daily usage ledger when a valid usage snapshot decreases.
- Continue crediting only positive usage deltas observed after the new baseline.
- Preserve the previous baseline when usage data is empty or unavailable.
- Add regression logging and tests for drop-and-recovery scenarios.

## Root cause

The companion ledger treated `todayTokens` as a monotonic counter and only processed values greater than `claimedTodayTokens`. Codex daily usage is recalculated from local rollout files and can decrease after replay, deduplication, or file updates. Once the snapshot dropped below the previous high-water mark, later usage was ignored and egg progress stopped.

## Behavior change

A valid lower snapshot becomes the new daily baseline without subtracting existing progress. Subsequent increases are credited from that baseline. Empty or unavailable usage data does not rebase the ledger, preventing a temporary zero from crediting the entire day again.

## Validation

- Local `swift build` passed with Xcode 26.6.
- Local `./scripts/test-gate.sh` passed: 519 tests executed, 7 skipped, 0 failures.
- Local logic-core line coverage passed: 89.29% (75% threshold).
- GitHub Actions `build-test` passed on `macos-15`.
- GitHub Actions `secret-scan` passed.

### DIFF
--- a/Sources/PokeTokenBar/Core/CompanionModel.swift
+++ b/Sources/PokeTokenBar/Core/CompanionModel.swift
@@ -504,7 +504,14 @@ struct CompanionState: Codable, Sendable {
     var eggTier: Rarity?
     // 알 상태에서 미리 롤해둔 부화 종(프리패칭) — 부화 순간 네트워크 딜레이 제거. 재시작에도 유지.
     var pendingHatchID: Int?
-    var claimedTodayTokens = 0
+    /// 오늘 사용량 적립 기준값 — 프로바이더별로 독립 관리한다.
+    ///
+    /// `nil`은 aggregate `claimedTodayTokens`만 가지고 있던 구버전 세이브가 아직 첫 유효
+    /// snapshot을 기준으로 seed되지 않았다는 뜻이다. 첫 update에서 현재 프로바이더 값을
+    /// 기준값으로만 저장하고, 과거 사용량은 소급 지급하지 않는다. 빈 map은 이미 seed된 뒤
+    /// 오늘 보고한 프로바이더가 없는 정상 상태와 구분되어야 하므로 `nil`과 별도로 유지한다.
+    /// 키는 `UsageProvider.id`를 그대로 사용한다.
+    var claimedTodayTokensByProvider: [String: Int]? = nil
     var lastDate = ""
     // 현재 포켓몬(없으면 알)
     var active: MonState?
@@ -534,7 +541,15 @@ struct CompanionState: Codable, Sendable {
         // 모르는 rawValue 는 nil(보증 없음)로 강등 — 관대 디코딩의 안전한 방향(있지도 않은 보증을 만들지 않는다).
         eggTier            = c.lenientOptional(Rarity.self, forKey: .eggTier)
         pendingHatchID     = c.lenientOptional(Int.self, forKey: .pendingHatchID)
-        claimedTodayTokens = c.lenient(Int.self, forKey: .claimedTodayTokens, default: 0)
+        if c.contains(.claimedTodayTokensByProvider) {
+            claimedTodayTokensByProvider = c.lenient([String: Int].self,
+                                                      forKey: .claimedTodayTokensByProvider,
+                                                      default: [:])
+        } else {
+            // 구버전의 aggregate claimedTodayTokens 키는 의도적으로 읽지 않는다. 프로바이더별
+            // 분해가 불가능하므로 다음 CompanionStore.update에서 현재 snapshot을 기준점으로 seed한다.
+            claimedTodayTokensByProvider = nil
+        }
         lastDate           = c.lenient(String.self, forKey: .lastDate, default: "")
         // active 손상(빈 pathIDs 등) → 알로 폴백하되 도감·인벤토리는 보존.
         active             = c.lenientOptional(MonState.self, forKey: .active)
--- a/Sources/PokeTokenBar/Core/CompanionStore.swift
+++ b/Sources/PokeTokenBar/Core/CompanionStore.swift
@@ -226,8 +226,9 @@ final class CompanionStore {
 
     // MARK: 갱신 (AppDelegate 가 UsageStore 값으로 호출)
 
-    func update(todayTokens: Int, todayDate: String, monthTotal: Int,
+    func update(todayTokensByProvider: [String: Int], todayDate: String, monthTotal: Int,
                 burnTier: BurnTier, limitWarning: Bool, hasUsageData: Bool) {
+        let todayTokens = todayTokensByProvider.values.reduce(0, +)
         if !state.installBaselineSet {
             // 설치 기준선 — 실제 데이터가 도착한 시점의 today 를 baseline 으로(이전 사용량 미카운트).
             // 데이터 도착 전(기동 직후 빈 새로고침)에는 잡지 않는다.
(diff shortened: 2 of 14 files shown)


chattymin:
Thanks for this — the diagnosis is correct and the write-up made it easy to verify. `claimedTodayTokens` is a high-water mark, so a downward recount freezes credit until usage climbs back past the old peak. I reproduced that on `main`: credit 1,000 → snapshot drops to 200 → 700 real tokens later, still 1,000 credited. And #137 (merged today) bumps `codexParserVersion`, which clears and re-parses the Codex cache — so upgrading users will hit exactly this.

The patch as written trades that for the opposite defect, because `hasUsageData` isn't "today's number is trustworthy" — it's `!snapshots.isEmpty` (`UsageStore.swift:144`). The total is a per-provider sum (`:140`), and a provider that succeeds with no entries for today is dropped from `snapshots` (`:529`, `:532`), so a partial loss passes the guard.

Two scenarios, run against this branch and `main` with stubbed providers:

1. claude 1,000 + codex 500 → codex returns nil (not a throw) → total 1,000 with `hasUsageData` still true → codex returns → **2,000 credited for 1,500 real tokens** (main: 1,500).
2. phase 2 appends a carrier snapshot with `today: nil` when a 5h block is active (`:563–570`), so `hasUsageData == true` while `todayTotalTokens == 0`. The ledger rebases to 0 → **4,000 credited for 2,000** — the case the new comment says is prevented. `SaveTransfer.swift:187–192` guards the same hazard on the import path.

```swift
// (1) production precondition: a partial loss keeps hasUsageData true
let claude = FakeProvider(id: "claude_code", daily: today(1_000))
let codex  = FakeProvider(id: "codex",       daily: today(500))
await store.refresh(scheduleEmptyRetry: false)   // total 1500, hasUsageData true
codex.daily = nil                                // not a throw — "no entries for today"
await store.refresh(scheduleEmptyRetry: false)   // total 1000, hasUsageData **true**

// (2) feeding that sequence to the companion
feed(0); feed(1_500); feed(1_000); feed(1_500)
// this branch: eggUsage = 2000 for 1500 real tokens
// main:        eggUsage = 1500
```

Also, a lower snapshot isn't necessarily a valid one: parse failures are swallowed (`LocalUsageReader.swift:889`) and blobs are cached by `(mtime, size)` (`LocalUsageCache.swift:11`), so a file read mid-write yields zero entries until that file changes again, then recovers — drop-then-recovery, i.e. phantom credit.

The new tests only cover `hasUsageData: false` (all providers gone); the branch that actually fires (`todayTokens < claimed && hasUsageData == true`) isn't exercised with a partial drop.

Suggested direction: keep the ledger per provider — credit `Σ max(0, currentᵢ − claimedᵢ)`, leave a provider that reports nothing untouched, and rebase only the line of a provider that reports a lower number. That removes the dead zone without re-crediting recovered totals, and stays provider-agnostic (a keyed map, no id literals). Touch points: the three `update` call sites (`PokeTokenBarApp.swift:147`, `FloatingPetPanel.swift:169`, `SettingsView.swift:515`), a new field on `CompanionState` with a seed migration, and the device-ledger set in `SaveTransferTests.swift:482`.

Happy either way — take it on if you'd like, or I can pick up the per-provider version and credit this PR. Either way, let's land both scenarios above as regression tests.

@ite476 — which would you prefer?

ite476:
@chattymin 

I appreciate your quick review. Since English is not my first language, the text below was AI-generated and may sound unnatural in places. Please focus on the technical meaning rather than the wording.

## Main topics for discussion

### 1. Introducing a per-provider ledger

I agree that keeping a separate ledger for each provider is a reasonable direction.

My initial patch treated the aggregate usage total as an independent and reliable source because I did not yet have a complete understanding of the provider snapshot lifecycle and the meaning of `hasUsageData` in the existing codebase.

Before introducing a provider-keyed ledger, however, I think we should check whether the codebase already has a strategy for stable and collision-free provider identifiers. For example, related sources such as `codex` and `codex-cli` may need to remain distinguishable.

If this is not already clearly defined, an ADR, a provider identifier naming policy, and additional comments around the relevant code may be useful.

Provider additions may not happen frequently, so this may not require a large amount of additional machinery. I am not familiar enough with the broader AI provider ecosystem to judge how common this risk is, so I would appreciate your opinion.

### 2. Additional edge-case concerns

I have a few additional edge cases in mind.

- A temporary error causes a provider to report `0` or a lower value.
- Usage continues normally, and the increments after the error are credited.
- The value from before the error later recovers.
- The recovered usage is then treated as new usage, which could effectively credit the same tokens twice.

```mermaid
flowchart LR
    A["Valid snapshot<br/>reported: 1,000<br/>already credited: 1,000"]
    B["Temporary error<br/>reported: 0<br/>ledger rebased to 0"]
    C["Usage continues<br/>reported: 100<br/>+100 credited"]
    D["More usage<br/>reported: 200<br/>+100 credited"]
    E["Previous value recovers<br/>reported: 1,000<br/>+800 credited again"]

    A --> B --> C --> D --> E

    classDef risk fill:#ffe5e5,stroke:#c62828,color:#111
    class B,E risk
```

The concern is that the original 1,000 tokens were already credited before the temporary drop. If the ledger is rebased to zero, the recovery to 1,000 can cause the same usage to be credited again, potentially resulting in duplicate crediting of most or all of the previously observed usage.

A similar issue could occur if a user switches between accounts within the same provider, such as Codex. The token context or usage reported under the same provider key could fluctuate between accounts.

However, I am not sure whether these concerns are necessary to address as part of the current fix, or whether they are overly cautious assumptions.

If you think these scenarios are unlikely under the current architecture or should be considered YAGNI, I am fine with excluding them from the current PR. We could record them as follow-up issues if they are worth tracking.

### 3. Parsing failures and partially written files

I agree that parsing failures and reads that occur while a file is being written should be considered.

I do not yet understand the current implementation well enough to propose the exact mechanism. If a provider snapshot is incomplete, invalid, or still being produced, we probably should not use it to update that provider’s ledger. Instead, we could preserve the last valid baseline and only update the ledger when a complete and valid snapshot is available.

This may be better expressed as an explicit snapshot state, such as:

- a valid and complete snapshot;
- a temporarily unavailable snapshot;
- an invalid or incomplete snapshot caused by a parsing or file-read problem.

However, if reliably detecting these states would make the implementation significantly more complex than the current evidence justifies, I would be inclined to treat this as YAGNI for now, document the assumption, and create a follow-up issue only if necessary.

### 4. Migration of existing saves

I agree that the migration policy for existing save data needs to be made explicit.

Since the current state stores an aggregate high-water mark rather than provider-specific values, it may not be possible to reconstruct the old provider-level ledger perfectly.

I have not yet inspected the current save and migration implementation in enough detail to give a definitive answer. Rather than making an assumption based on incomplete information, would you be willing to propose the migration approach that best fits the current implementation? I can then review it and help implement or test the agreed approach.

In particular, we should decide:

- how to seed the provider-level ledger from the existing aggregate high-water mark;
- how to preserve already credited egg or Pokémon progress;
- how to prevent historical usage from being credited again after migration;
- which snapshot should become the new baseline after migration.

### 5. Regression tests

I agree that the scenarios mentioned above should be added as regression tests.

If possible, I would also like to cover:

- partial loss of one provider’s data;
- a carrier snapshot with `today == nil`;
- a temporary zero or lower value followed by recovery;
- a genuine decrease in one provider’s valid snapshot followed by new usage;
- migration from the existing aggregate ledger to the provider-level ledger.

If some tests later turn out to be too fragile, I would prefer to rewrite them around observable behavior and the intended contract rather than simply removing them.

If a scenario is ultimately considered outside the product’s guarantees or YAGNI, we can record that decision in an issue and close it without implementing it.

## Implementation direction and work division

This PR is closer to an exploratory draft than to a merge-ready implementation. I found and reproduced the problem, then tried an initial fix based on my understanding of the codebase.

The provider identity, snapshot validity, ledger structure, and save migration concerns seem closely connected. They also affect the long-term direction of the repository. Since I do not yet have a complete understanding of the existing design and policies, I am slightly concerned about making those structural decisions independently.

For that reason, I think direct decisions about the repository’s overall direction would be safer if they were led by you as the reviewer or maintainer.

If you identify concrete and separable tasks, I can take responsibility for implementing those pieces and adding the relevant regression tests. On the other hand, if the changes are tightly coupled and cannot be separated cleanly, I would be comfortable with you leading the overall implementation to reduce the risk of conflicting with the existing codebase.

This is not intended to hand off responsibility. My intention is to contribute the initial issue discovery, reproduction, and implementation work while respecting the repository context and the maintainer’s architectural direction.

## Closing thoughts

For work that is directly related to and required for this fix, GitHub Stacked PRs may be useful if each PR can be reviewed independently and remain buildable and testable.

For related concerns that are not required for the current fix, such as account-aware provider identity, a separate GitHub Issue may be more appropriate so that they remain explicitly outside the current scope.

I know these are possible approaches, but I do not know this repository’s preferred operating model well enough to decide which one is most appropriate. Please feel free to choose between stacked PRs, follow-up issues, or another workflow based on the repository’s conventions.

Before deciding who should implement the per-provider version, I would like to agree on the provider identity policy, snapshot validity assumptions, migration scope, and the boundaries of the current fix.

chattymin:
@ite476 Thank you for the deep thought you put into this and for giving me the chance to reflect on the direction — thanks to you, I get to work toward a better product.

Direction agreed. Decisions first, then the plan: two PRs. The fix itself stays on this branch (keeps your authorship); the torn-read guard follows as a separate PR so it can be reviewed and, if ever needed, reverted independently.

**Decisions**

- **Provider identity**: already settled — one implementation, one constant id per source, registered only in `UsageStore.init`. `docs/reference/provider-extension.md` is the policy; the ledger keys on `UsageProvider.id` as-is. No ADR needed.
- **Drop-to-zero flow**: mostly can't happen. A throw keeps the previous value, an empty success drops the provider from `snapshots`, the carrier has `today == nil` — all read as "reports nothing" → line untouched. The only real hazard is a mid-write read yielding a lower non-zero count, then recovering; the follow-up PR guards that. No state machine needed.
- **Account switching**: rollout files are local and don't reset on switch — YAGNI, no tracking issue.

**This PR — per-provider ledger + migration**

- Plumbing: add `todayTokensByProvider: [String: Int]` to `UsageStore` (same date guard as `todayTotalTokens`) and pass it into `CompanionStore.update` and `applySave`. Parity test: map sum == aggregate.
- Ledger: replace `claimedTodayTokens` with a `[String: Int]` map. Credit `Σ max(0, current_i − claimed_i)`; absent provider → untouched; lower value → rebase that line only.
- Migration: first update seeds `claimed_i = current_i`, then drop the old aggregate field. Classify the new field in the device-ledger set (`SaveTransferTests` forces this) and rebase imported lines like `SaveTransfer.rebasedForThisDevice`.
- Regression tests: my two review scenarios, per-provider decrease + new usage, migration seed, import rebase.

**Follow-up PR — torn-read guard**

- Rebase downward only after the same lower value is seen on two consecutive updates (in-memory pending slot, reset on any different value). Tests: drop-then-recovery with the guard holding and firing.

**Ground rules**: both PRs pass `scripts/test-gate.sh` before review; I'll review promptly. Out of scope: account-aware identity, snapshot state machine, ADR. If the code contradicts anything above, flag it on the PR instead of working around it.

ite476:
@chattymin 

Thanks for the clear direction. I implemented the agreed scope for this PR.

Completed:

- Added provider-keyed daily usage tracking using `UsageProvider.id`.
- Replaced the aggregate companion ledger with a per-provider ledger.
- Preserved missing provider lines during same-day partial snapshots.
- Rebases only the affected provider line when a lower value is observed.
- Added legacy aggregate-save migration and import baseline handling.
- Added regression tests for partial provider loss, carrier snapshots, valid decreases, stale imports, and late provider recovery across a date boundary.

Validation:

- `scripts/test-gate.sh`
- 587 tests executed, 7 skipped, 0 failures
- Logic-core line coverage: 89.61%
- CI passed for commit `aeeacf0`

The torn-read guard remains intentionally scoped to the follow-up PR. Account-aware provider identity, a snapshot state machine, and an ADR also remain out of scope as agreed.

The branch is ready for review.

chattymin:

PR #147, opened 2026-08-06
feat: split the collection tab into a species Pokédex and a catch log

parkjh2568:
## Summary

**Proposing a direction — I'd like your call before this goes any further.**

The Collection tab is labelled Pokédex, but it stores one row per individual:
`graduate()` appends a fresh `DexEntry` every time, so raising the same evolution
line twice draws its chain as two identical rows. My save has Yamask → Cofagrigus
twice, differing only in nature. It already behaves as a catch log, and the
Pokédex name promises something it doesn't do.

So I kept the log and added the missing half, as two screens behind a segmented
control in the same tab (top-level tabs stay at four):

- **Catch log** — the existing list, unchanged in what it shows. Sorted by catch
  time now, which is what the name implies. Rarity used to be the primary sort
  key, so a fresh graduation sat below older-but-rarer catches; narrowing by
  rarity is what the filter capsules are for.
- **Pokédex** — species-level. One cell per species you've owned, dex-number
  order, 24 per page (4x6). Duplicates fold into one cell, and the natures you
  caught that species with become tags. Selecting a cell lists every nature in the
  footer line. A ✨ beside the dex number marks species you own shiny, and
  selecting it swaps the cell to shiny colours.

Everything is derived from existing state, so there's **no schema change, no
migration, and no network calls**.

### Questions for you

1. Is splitting the tab the direction you want, or would you rather the current
   screen just deduplicate in place?
2. The Pokédex only shows what's caught — no silhouettes for unowned species.
   Reaching 649 would take years at real token rates, so a mostly-empty grid felt
   worse than showing only what you have. Fine, or do you want the full-dex
   framing?
3. Shiny is a per-species flag, so a folded cell can't say *which* individual was
   shiny. Marking the species and revealing the colour on selection was my
   compromise — the catch log still pairs shiny with nature per individual.
4. The collection tab is 580pt tall while Shop/Bag are 520pt. 4x6 with a 40pt
   sprite plus name and tag doesn't fit in 520 once the segment takes a row. I can
   shrink the sprite and go to 520 if you prefer uniform tab heights.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| One tab titled Pokédex (도감 / 図鑑). One row per Pokémon raised, with its full evolution line, rarity, nature and capture date. Sorted rarity-first, so the same line raised N times produced N identical rows, and a new common graduation sat below older uncommons. | Two screens behind a `Pokédex / Catch log` segment. **Catch log** is the old list, newest first. **Pokédex** is a 4x6 grid of owned species in dex-number order — one cell per species, nature tags, a `×N` badge only when the tag count can't account for the number obtained, a ✨ beside the dex number for shiny, and a footer that lists every nature for the selected cell. |

Screenshots: `assets/screenshot-collection-pokedex.png` (a cell selected, showing
the footer) and `assets/screenshot-collection-catchlog.png`, both referenced from
the three READMEs. They replace `assets/screenshot-collection.png`, which showed
the old single screen.

Two existing defects the new screens surfaced, both fixed here because the
Pokédex depends on them:

- `SpriteView`'s reload guard compared only `loadedID`, so flipping `shiny` on the
  same species re-ran `.task` and then bailed out as already-loaded — the shiny
  toggle would have done nothing. Ditto's reveal changes species and shiny
  together, which is why nothing had hit it before.
- `Text(_, style: .relative)` followed `Locale.current` while `L` strings follow
  `AppLanguage`, so a Korean-locale Mac running the app in English rendered "Catch
  log" beside "3시간 46분". Sorting the log by time put that timestamp at the
  centre of the screen, which is how I noticed. The popover root now hands the app
  locale down, fixing all eight sites — seven of them in the limits section, which
  is why this PR touches that file.

Also: `SpriteStore.memLimit` 24 → 64 (a full page is exactly 24 sprites, so the
LRU rotated completely on every page turn), the Korean tab label becomes 컬렉션
(도감 now collides with the segment inside it), and ✨ carries a localized shiny
noun as an accessibility label on both screens.

## Notes

- `Rarity.sortRank` stays. It's no longer a sort key anywhere, but the premium egg
  guarantee compares two rarities with it, so its doc comment now names that
  consumer instead of the Pokédex sort that no longer exists.
- `docs/reference/defect-log.md` gains the two class rules these bugs belong to.
- One thing I couldn't verify by eye: a cell is 80pt wide, and a Korean nature name
  plus both badges is about 4pt over. `lineLimit(1)` with `minimumScaleFactor`
  shrinks it rather than clipping, so it can't inflate the popover, but no save I
  have produces that combination.

## Checklist

- [x] `swift build` and `swift test` pass locally (`./scripts/test-gate.sh` green,
      547 tests, logic-core line coverage 89.67%)


chattymin:
@parkjh2568 Thank you for your interest in this project and for proposing a new feature. Answers to your questions:

1. **Split**: yes, keep it as implemented.
2. **Owned-only**: agreed — no silhouettes.
3. **Shiny flag + tap reveal**: agreed.
4. **Height**: see the change below — it may let the grid fit 520 again. If it does, go back to 520; if not, 580 is fine.

One change request: **drop natures from the Pokédex screen entirely**. The Pokédex should show species only — natures and per-individual info (including the ×N count) belong to the catch log. The footer's existing rarity fallback is enough for a selected cell, and keep the shiny tap-to-reveal as is.

Everything else (both defect fixes, the sort change, the docs) looks good. Happy to merge once that's in.

parkjh2568:
@chattymin Thanks — all of it is in as of 40a81bc.

- **Pokédex is species-only.** `DexSpecies` is `(id, name, rarity, isShiny)`; a cell is
  number + sprite + name, and selecting one puts `#554 Darumaka · Uncommon` in the footer.
  Natures, `obtained` counts and the `×N` badge are gone.
- **Height back to 520.** Dropping the tag row freed exactly what had forced 580 — the
  sprite even grew 40 → 44 with 3.8pt spare. No popover resize on any tab or segment.
- **Shiny tap-to-reveal unchanged**, including the Ditto rule (disguised hides it).
- **`×N` left out on purpose.** Branching lines have no single denominator — base-keyed,
  final-keyed and chain-keyed counts disagree for a split Eevee, and a log row shows the
  whole chain either way. Better added once the rule is one you're happy with, not guessed
  now; glad to take it separately.
- **One test needed a new observable.** Without `obtained`, the `plannedPathIDs` guard had
  nothing to watch, so it now injects an active with an empty dex: `[1]` correct, `[1,2]`
  if `pathIDs` is used whole, `[1,2,3]` if the planned route is. Verified by injecting
  each defect.

Screenshots and the three READMEs are regenerated. `./scripts/test-gate.sh` green, 597
tests, 0 failures, coverage 89.67%.

chattymin:

PR #155, opened 2026-08-11
feat: track GitHub Copilot CLI usage

gangpro:
## Summary

GitHub Copilot CLI records every assistant request in its own local session store
(`~/.copilot/session-store.db`, table `assistant_usage_events`), but no provider read it,
so those tokens never reached the companion or the usage tracker.

This adds `LocalCopilotProvider` on top of the existing additional-source cache
(`LocalAdditionalUsageProvider.swift`) and registers it in the default `providers:` array of
`UsageStore.init` — the two extension points named in CONTRIBUTING. Daily / 5h block / weekly /
monthly all come from the shared `LocalUsageReader` aggregation, so no generic path branches on
the new provider id.

### What the format required

- **`input_tokens` is the whole prompt.** `cache_read_tokens` and `cache_write_tokens` are a
  *subset* of it, so the reader subtracts them. Summing the columns as spelled would have
  counted a cached prompt three times — the same "same spelling, different meaning" trap the
  defect log records for Grok. Verified against a real store: a row with
  `input_tokens=43792, cache_read=42698, cache_write=904` bills 190 uncached prompt tokens,
  which matches the row's own `token_details_json` breakdown.
- **`reasoning_tokens` is a breakdown of `output_tokens`**, not an extra charge, so it is not
  added on top.
- **Every row is one billed API call**, including rows a subagent produced. Unlike Grok's
  session logs, subagent usage is not folded into the parent turn, so nothing is excluded.
- **`created_at` has two shapes**: the CLI writes ISO-8601 with a `Z`, while the column default
  (`datetime('now')`) writes `YYYY-MM-DD HH:MM:SS` in UTC. Both are parsed, offsets included.
- **The table is append-only with an AUTOINCREMENT id**, so refreshes replay only rows past the
  previous high-water id. A max id below the watermark means the store was pruned or recreated
  and forces a cold rescan, matching the Cursor watermark handling.

### Two hazards the incremental read has to respect

A row skipped once is never revisited, because a later id advances the watermark past it. Both
of these were found in review of the first push and are covered by regression tests:

- **The SQL cutoff on `created_at` is a lexicographic compare, not a time compare.** A same-day
  ISO row is safe (its `T` sorts after the cutoff's space separator), but a negative UTC offset
  renders an earlier calendar day than the instant it represents —
  `SELECT '2026-01-03T20:00:00-05:00' >= '2026-01-04 00:00:00'` is `0`, even though that row is
  `2026-01-04T01:00:00Z`. The cutoff is therefore backed off a full day, which covers every
  offset SQLite accepts (±14h); the exact filter stays in Swift.
- **Row ids are unique only within one store**, and `$COPILOT_HOME` may name several, each
  numbering from 1. Entry ids are namespaced by database path so dedup cannot collapse id 1 of
  one store into id 1 of another.

Copilot bills subscription premium requests rather than per-token dollars, so the provider
reports tokens only (`reportsCost = false`), like Cursor and Antigravity.

Reads are read-only SQLite; `$COPILOT_HOME` is honoured for a relocated Copilot home.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

No files under `Sources/PokeTokenBar/UI/` changed. The only visible effect is that a
"Copilot" tab now appears in the existing per-service tab bar once Copilot CLI usage is
detected, and its tokens join the combined totals. `ProviderTabLayoutTests` was updated so the
tab-bar overflow guards measure the registered set including the new tab.

## Verification

- `swift build` passes; CI is green on the previous push of this branch (all 14 new tests
  passed there).
- `swift test` could **not** be run on this machine — only the Command Line Tools are
  installed, so `XCTest` is unavailable (`no such module 'XCTest'`). Relying on CI for the
  suite; happy to iterate if anything fails.
- Because unit tests alone would only prove the parser agrees with my own fixtures, the
  provider was also exercised end-to-end through a scratch harness against a **frozen copy of a
  real `~/.copilot/session-store.db`** (1980 rows), going through `LocalCopilotProvider` →
  `LocalAdditionalUsageCache` → `LocalUsageReader` exactly as the app does:

  | | provider | independent SQL over the same frozen DB |
  |---|---|---|
  | today | 15,329,978 | 15,329,978 |
  | week | 22,483,451 | 22,483,451 |
  | month | 49,546,948 | 49,546,948 |

  So the cache subtraction reproduces the source of truth exactly rather than by accident, and
  the active 5h block, week and month windows populate.
- The branch-level behaviour the tests assert was also run against synthetic stores: both
  timestamp shapes, a negative UTC offset row, zero-token rows skipped, the day-boundary
  cutoff, an incremental read returning only new rows, an empty incremental read preserving the
  watermark, a store deleted and recreated with restarted ids, two stores sharing row ids, a
  direct `.db` path as root, and a nonexistent root.

## Checklist

- [x] `swift build` passes locally (`swift test` needs Xcode — see Verification)

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -17,9 +17,9 @@
 
 </div>
 
-PokeTokenBar は、あなたがすでに使っている AI コーディングトークン（Claude Code・Codex・Gemini CLI・Antigravity・OpenCode・Hermes Agent・Cursor・Grok CLI）を、macOS メニューバーの中で育っていく **ポケモンのパートナー** に変えます。トークンを使うとタマゴが孵化し、実際の進化ラインに沿って進化し、最終進化後に図鑑へ卒業して、また新しいタマゴが始まります。パートナーの下には正確な使用量トラッカーがあります — 今日の使用量・コスト、公式の5時間／週間上限をローカルログから直接読み取ります。
+PokeTokenBar は、あなたがすでに使っている AI コーディングトークン（Claude Code・Codex・Gemini CLI・Antigravity・OpenCode・Hermes Agent・Cursor・Grok CLI・Copilot CLI）を、macOS メニューバーの中で育っていく **ポケモンのパートナー** に変えます。トークンを使うとタマゴが孵化し、実際の進化ラインに沿って進化し、最終進化後に図鑑へ卒業して、また新しいタマゴが始まります。パートナーの下には正確な使用量トラッカーがあります — 今日の使用量・コスト、公式の5時間／週間上限をローカルログから直接読み取ります。
 
-> トークン使用量はローカルの Claude Code・Codex・Gemini CLI・Antigravity・OpenCode・Hermes Agent・Cursor・Grok CLI データから直接読み取ります（`totalTokens` = input + output + cache、ローカル日付）— 外部 CLI 不要。非公式・非商用のポケモンファンプロジェクトです — [ライセンス & 免責](#ライセンス--免責) を参照。
+> トークン使用量はローカルの Claude Code・Codex・Gemini CLI・Antigravity・OpenCode・Hermes Agent・Cursor・Grok CLI・Copilot CLI データから直接読み取ります（`totalTokens` = input + output + cache、ローカル日付）— 外部 CLI 不要。非公式・非商用のポケモンファンプロジェクトです — [ライセンス & 免責](#ライセンス--免責) を参照。
 
 ## なぜ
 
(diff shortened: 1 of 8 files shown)


gangpro:
### Why I opened this

I use Copilot CLI as my daily driver, and I noticed PokeTokenBar was not collecting any of its
tokens — the egg simply never moved on the days I only used Copilot. Every other CLI I use is
tracked, so rather than file an issue I tried adding the provider myself. (I am new to this
codebase, so please tell me if anything here is off — I am happy to rework it.)

### How you can verify it yourself

The tricky part of this format is that **`input_tokens` already contains the cached prompt**
(`cache_read_tokens` + `cache_write_tokens` are a subset of it), so the correct token total for
a row is simply `input_tokens + output_tokens`. That gives an independent cross-check that does
not go through any of my code:

```bash
# 1. unit tests for the parser
swift test --filter CopilotUsageTests

# 2. what today's Copilot total should be, computed straight from your own store
#    (assumes your machine's local timezone; needs some Copilot CLI usage today)
sqlite3 ~/.copilot/session-store.db "
  SELECT sum(input_tokens + output_tokens)
  FROM assistant_usage_events
  WHERE created_at >= strftime('%Y-%m-%dT%H:%M:%SZ',
                               datetime('now','localtime','start of day','utc'));"

# 3. launch the app and compare the number on the Copilot tab
./scripts/build-app.sh && open /Applications/PokeTokenBar.app
```

Step 2 should equal what the Copilot tab shows for today.

### What I measured

I could not run `swift test` on my machine — I only have the Command Line Tools installed, so
`XCTest` is unavailable (`no such module 'XCTest'`). CI covers the suite, but since unit tests
only prove the parser agrees with my own fixtures, I also ran the provider against a **frozen
copy of my real `~/.copilot/session-store.db`** (1980 rows) through
`LocalCopilotProvider` → `LocalAdditionalUsageCache` → `LocalUsageReader`, i.e. the same path
the app takes, and compared it with the SQL above over the same frozen file:

| window | provider | independent SQL |
|---|---|---|
| today | 15,329,978 | 15,329,978 |
| this week | 22,483,451 | 22,483,451 |
| this month | 49,546,948 | 49,546,948 |

The active 5-hour block also populates, and a second call goes through the incremental branch
and returns the same total without rereading the store.

### One thing worth a second opinion

The SQL cutoff on `created_at` is a **lexicographic** compare, since the column is text. That
is subtle enough that I want to flag it rather than bury it:

```sql
sqlite> SELECT '2026-01-03T20:00:00-05:00' >= '2026-01-04 00:00:00';
0
```

That row is `2026-01-04T01:00:00Z`, so it belongs in the window, and because a later id would
advance the watermark past it, it would have been lost permanently. The cutoff is therefore
backed off a full day (covering every offset SQLite accepts) and the exact filter stays in
Swift. `testNegativeUTCOffsetRowIsNotDroppedByTheTextCutoff` pins this.

Similarly, row ids are unique only within one store while `$COPILOT_HOME` may name several, so
entry ids are namespaced by database path — otherwise dedup would collapse id 1 of each store
into one event (`testTwoStoresWithTheSameRowIDsAreBothCounted`).

chattymin:

PR #158, opened 2026-08-13
feat: allow pinning a representative Pokémon

dev-minsoo:
## Summary

This PR lets users pin any owned Pokédex species as their representative Pokémon for the menu bar and optional floating pet, independently of the companion they are currently raising.

Previously, both surfaces always followed the active companion. Separating these choices lets users continue raising any Pokémon while keeping a preferred owned species visible.

The representative Pokémon is configured from a single control under General settings, with two options: follow the current companion or choose a species in the Pokédex. The selection can also be changed from the Pokédex action or context menu and persists with the companion save.

While a representative Pokémon is pinned, the menu bar no longer follows egg, hatch, or evolution changes. Raising progress remains visible on Home.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| The menu bar and floating pet always followed the companion currently being raised. | Any owned Pokédex species can be pinned to the menu bar and optional floating pet. |
| There was no representative Pokémon control in General settings. | General settings provides one control with “Follow current companion” and “Choose in Pokédex…” options. |
| The Pokédex had selection, shiny, and raising states only. | The representative species has an accent-tinted background and a star capsule, and can be changed using the grid action or context menu. |

## Checklist


chattymin:
@dev-minsoo Thanks for tracing every path where the pinned species can vanish, and for updating the transfer-classification guard.

I want this feature in, so this is a request for changes rather than a rejection: two design changes, plus a rebase and one performance note.

**1. Settings: one control, two options** — Follow current companion / Choose in Pokédex…. Both strings are already in the PR. If the Pokédex keeps its own action, give it the same "follow" wording instead of "Unset pet", so there is one concept rather than two.

**2. Apply it to the menu bar, not just the pet.**
- Source: `PokeTokenBarApp.swift:189` reads `currentSpeciesID`/`currentIsShiny` — route it through the same subject.
- Add the field to `observeCompanionSprite()` (`:97`), or a change won't reach the menu bar until the next poll (120s default).
- Rename `floatingPet*` → `representative*`: it drives two surfaces now, and the field hasn't shipped, so this is the one point where renaming costs no save migration.
- Move the row from **Floating Pet** to **General** and drop the `floatingPetEnabled` gate — the menu bar is always on.
- README (×3) still describes it as pet-only.
- Product note: the menu bar sprite is the only always-visible signal of the companion being raised (egg → hatch → evolution), and pinning replaces it, egg included. It stays opt-in, so a line in the setting or README is enough.

**Rebase.** `Localization.swift` is the only conflict: `t()` takes `es` since #159, so your six new strings need Spanish.

**Perf.** `floatingPetSubject` folds `dexSpecies` on every render of an always-on view, and widens what that view observes to `dex`/`active`/`currentLine`. A species-level shiny lookup or a cached value avoids both — more so once the menu bar reads it too.

dev-minsoo:
Thanks for the detailed feedback. I’ve addressed all the requested changes and rebased the branch onto the latest `main`.

- Consolidated the representative Pokémon setting into a single control under General, with “Follow current companion” and “Choose in Pokédex…” options.
- Applied the representative Pokémon to both the menu bar and floating pet, including immediate menu bar updates when the selection changes.
- Renamed the feature internals from `floatingPet*` to `representative*` and removed the floating-pet-enabled dependency.
- Replaced the repeated Pokédex fold with a cached representative subject shared by the always-on surfaces.
- Added Spanish translations for the new strings and updated all three README files to explain the menu bar behavior and its effect on egg, hatch, and evolution changes.
- Updated and added tests for the revised behavior.

Both `build-test` and `secret-scan` are passing. I’d appreciate another review when you have time.

chattymin:
@dev-minsoo All six items from the last round are in. `swift build` / `swift test` pass (684 tests), and I ran throwaway probes on the state layer — follow-mode tracking through hatch and evolution, the Ditto reveal clearing a pin on the disguise species, and the Pokédex cell set matching the pinnable set. All green. Routing the menu bar through the cached subject held up, which was the part I was least sure about.

Three changes before merge.

**1. `Choose in Pokédex…` can land on the Catch log — needs a fix**

`openRepresentativeDex()` sets `tab = .collection`, but the Pokédex/Catch-log segment is `CollectionView.showingLog` — a `@State` that survives popover close (the note at `PopoverView.swift:16` spells this out), and `reset()` at `PokeTokenBarApp.swift:366` doesn't clear it. If the user last viewed the Catch log, the menu item lands there, and log rows carry no representative action. The doc comment claims it reuses the default Pokédex segment; nothing enforces that.

This predates your rebase — the same body and comment were in `openFloatingPetDex()`. It matters now because I asked you to move the row to General and drop the `floatingPetEnabled` gate, so every user can reach it rather than only floating-pet users.

Suggested fix: hoist the segment into `PopoverNavigation` — the pattern that file's header note already describes — and set it to the Pokédex in `openRepresentativeDex()`.

**2. The footer button truncates the line it sits on — my wording caused this**

I asked you to replace `Unset pet` with the same "follow" wording. That wording is wider than I accounted for. Intrinsic widths of the footer row against `PopoverMetrics.contentWidth` (332pt):

| state | before | now |
| --- | --- | --- |
| en, unpinned | `Set as pet` 268 | `Set as representative` 317 |
| en, pinned | `Unset pet` 266 | `Follow current companion` **337** |
| es, unpinned | — | `Establecer como representante` **361** |
| es, pinned | — | `Seguir al compañero actual` **343** |

Past 332 the species name absorbs the overflow, so the row renders `#0025 Pikachu · Co…` and `#0025 Pikach…`. `CompanionView.swift:881` says the rarity is precisely what selecting a cell buys you, and it is the first thing dropped.

Suggested fix: `.labelStyle(.iconOnly)` plus `.help(...)` on that button. I measured it at every locale and the name and rarity survive; the context menu you added already carries the wording.

**3. Regression test for the Ditto-reveal reconcile — recommended**

`reconcileRepresentativeSelection()` has two call sites no test reaches: the Ditto reveal (`CompanionStore.swift:1027`) and the evolution normalization (`:1050`). I confirmed the reveal path works, so this is a coverage gap rather than a bug — but the defect protocol in `CLAUDE.md` asks that a new guard be exercised by the branch that triggers it. Seed a disguised Ditto, pin the disguise species, drive the reveal, assert the pin clears and the subject falls back to Ditto.

dev-minsoo:
Thanks for the detailed follow-up — and for catching the locale-specific truncation. I’ve addressed all three points in a977ff4.

- `Choose in Pokédex…` now always opens the Pokédex, even when Catch Log was the last selected collection view.
- The footer action is now icon-only with localized help and accessibility labels, preventing the species name and rarity from truncating in English and Spanish. I also added a render regression test covering every supported language and both action states.
- Added a Ditto reveal regression test that pins the disguise species, reveals Ditto, and verifies that the pin is cleared and the representative falls back to the revealed shiny Ditto.

I verified all three flows locally. The full suite passes with 686 tests and 89.91% core coverage, and both debug and release builds pass.

chattymin:

Issue #157, opened 2026-08-13
The Cursor and Copilot readers duplicate the incremental SQLite scan, so a watermark fix has to land twice


chattymin:
`LocalAdditionalUsageProvider.swift` now carries two incremental SQLite readers with the same
shape: the Cursor one at `:337-556` and the Copilot one at `:557-748`. Normalising the provider
name and dropping comments, 96 of the Copilot section's 142 code lines are identical to the
Cursor section — 67%.

What is shared is the scaffolding, not the format:

- roots → per-database watermark map (`cursorWatermarks` / `copilotWatermarks`)
- `scanCursorRoots` `:397` / `scanCopilotRoots` `:619` — same loop, same accumulation, same
  `dedupKeepMax` + `date >= modifiedSince` filter
- shrink detection and the cold-rescan recovery
- the `nil`-vs-empty distinction that keeps a failed read from resetting a watermark

What actually differs per provider is about 46 lines: the database URL mapping, the `MAX(id)`
table name, the row query, and the row → `Entry` parse.

The cost is that a fix to the watermark logic now has to land in two places. Two lines carrying
the delicate part of that logic already exist twice, verbatim:

- `let didReset = afterRowID > 0 && maxRowID < afterRowID` — `:452` and `:658`
- `if highWater == 0 { highWater = effectiveAfter }` — `:485` and `:693`

Both encode a rule that is easy to get wrong in a way no crash reports: the second one is the
difference between "this incremental read found nothing new" and "reset the watermark to zero
and replay the month". A regression fixed in one copy would sit untouched in the other.

Not urgent, and deliberately not folded into the change that introduced the second copy —
factoring it out means editing the shipped Cursor reader, which is a wider blast radius than the
addition itself warranted. Rule of three: the right moment is the third incremental SQLite
provider. At that point the shape to extract is roughly

```
scanIncrementalStores(
  roots:, modifiedSince:, marks:,
  databaseURL: (URL) -> URL,
  maxRowIDSQL: String,
  rowSQL: String,
  parse: (OpaquePointer, URL) -> Entry?
) -> LoadResult
```

with `CursorLoadResult` / `CopilotLoadResult` collapsing into one type, since they already have
identical members.

Worth doing together with #152, which touches the same failure paths in the same file.

shubhransh-gupta:
Hi — I opened a PR for this from my fork. Happy to iterate on any review feedback.

Issue #152, opened 2026-08-08
LocalAdditionalUsageReader.query swallows every read failure, so OpenCode/Hermes/Cursor cannot be diagnosed


2JIHAN:
Found while working on #141. `LocalAntigravityUsageReader` had the same shape and #151 fixes it there; this is the sibling reader, which OpenCode, Hermes and Cursor all share.

`LocalAdditionalUsageProvider.swift` folds all three failure paths into the same `nil`:

- `:619-621` — `sqlite3_open_v2` fails → `return nil`
- `:625-626` — `sqlite3_prepare_v2` fails → `return nil`
- `:641` — `guard step == SQLITE_DONE else { return nil }`

The last one is verbatim the case raised in review on #141: dropping a partial read is right, but `nil` is indistinguishable from "this database has no rows", and nothing is written anywhere. A store that is persistently busy or unreadable reads as no usage for as long as it stays that way, with no way to tell from the outside.

A missing table is *not* the same class and should stay silent — `SQLITE_ERROR` from `prepare` is a permanent property of the file, and naming it would repeat every refresh for the life of the install. #151 splits on exactly that boundary.

Two constraints worth carrying over from #151:

- The decision has to be a pure value rather than an `AppLog.write` at the failure site. `AppLog` returns early outside the bundled app (`AppEnv.isBundledApp`), so a test that watched the log file would cover nothing — the same reason the limit-alert decision was split out of its own side effect.
- The volume has to be bounded. An unreadable source is re-read on every refresh — 720 times a day at the default interval — so one line per source per scan can rotate the 2 MB log and take the crash history in that same file with it.

Separately, `:622` is `defer { sqlite3_close(database) }` — the same call as note 4 on #141, safe today for the same accidental reason, that the `finalize` is registered afterwards and `defer` runs last-in-first-out.

chattymin:
closed : https://github.com/chattymin/PokeTokenBar/pull/151

PR #160, opened 2026-08-13
refactor: share Cursor and Copilot incremental SQLite scan (#157)

Borisserz:
## Summary

Cursor and Copilot each carried a full incremental SQLite scanner. Normalising names, 96 of Copilot's 142 code lines matched Cursor — including the two watermark lines that are easy to get wrong without a crash:

- `let didReset = afterRowID > 0 && maxRowID < afterRowID`
- `if highWater == 0 { highWater = effectiveAfter }`

A fix in one copy would sit untouched in the other. This extracts `scanIncrementalStores` as sketched in #157: URL mapping, `MAX` SQL, row query, and parse stay with the provider; the watermark loop exists once. `CursorLoadResult` / `CopilotLoadResult` collapse to `IncrementalStoreLoadResult`.

`rowSQL` is a closure rather than a single string. Cursor has two statements (cold `GLOB` vs incremental `NOT INDEXED`), and Copilot binds a second text cutoff (`created_at >= ?2`). Pinning a shared string would have changed one of those paths. `cursorIncrementalQueryPlan` now calls the same `cursorRowQuery` the scanner uses, so dropping `NOT INDEXED` fails the EXPLAIN test.

#152's diagnostic split for `query()` is not in this PR. `query()` already returns `nil` unless `sqlite3_step` ends on `SQLITE_DONE`; turning that into a logged value type also touches OpenCode/Hermes and is a wider blast radius than #157.

## Type of change

- [x] Refactor / cleanup

## Checklist

No UI changes.

## Tests

Shared helper on a format-neutral table (`IncrementalStoreScanTests`):

- `testFailedMaxIsNotAShrink` — corrupt file keeps the watermark, `didReset == false`
- `testEmptyIncrementalReadPreservesWatermark` — `highWater == 0` keeps `effectiveAfter`
- `testShrinkInOneRootColdRescansEveryRoot` — reset in B still returns A's history **and** A's watermark

Cursor-only (existing + empty incremental + healthy-root watermark on dual-root reset). Copilot-only, the path that previously had no copy of the #125 hardening tests:

- `testUnreadableDatabaseKeepsWatermarkAndDoesNotReset`
- `testResetInOneRootKeepsOtherRootHistory` (entries + healthy-store watermark)

`swift test --filter 'IncrementalStoreScanTests|CursorUsageTests|CopilotUsageTests'`: 43 tests, 0 failures.

Full suite: 624 pass, 9 skipped. The remaining failure is `LocalUsageCacheTests.testFormatterEdges` (`253 412 890` vs `253,412,890`) — locale grouping, unrelated to this change, TokenFormatter not touched.

## How to verify

```bash
  swift test --filter 'IncrementalStoreScanTests|CursorUsageTests|CopilotUsageTests'
```

### DIFF
--- a/CONTRIBUTING.ja.md
+++ b/CONTRIBUTING.ja.md
@@ -70,6 +70,9 @@ CI はすべてのプルリクエストで `swift build` と `swift test` を実
 - **バージョンマネージャー / インストールパスの追加** =
   `BinaryLocator.commonToolDirectories()` に追加します — 探索と子プロセスの `PATH` が
   共有する単一のソースです。
+- **追記専用 SQLite 使用量ストアの追加**（Cursor・Copilot のように rowid/`id` ウォーターマーク）=
+  `LocalAdditionalUsageReader.scanIncrementalStores` に URL / `MAX` SQL / row query /
+  parse だけ渡す。ウォーターマーク・ループをコピーしない。
 
 ## 法務 / 知的財産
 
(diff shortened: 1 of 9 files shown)


chattymin:

PR #159, opened 2026-08-13
feat: add Spanish (es) UI language

sonms:
## Summary
Adds Spanish (`es`) as a fourth UI language, following the architecture guide in #149 

* `AppLanguage` gains a `.es` case (`apiCodes: ["es"]`, native label `"Español"`), and `systemDefault` maps the `es` system locale to it.
* `L.t(...)` is extended to four languages; every UI string in `Localization.swift` (popover, settings, notifications, floating pet, shop, save transfer, Pokédex/catch log) is translated — the compiler's exhaustiveness check confirms nothing was left as an English placeholder.
* `PokemonNature.name(_:)` returns the 25 official Spanish nature names, cross-checked against PokéAPI's `names` data per nature rather than invented (this also fixed gender-agreement issues in the nature adjectives — `naturaleza` is feminine).
* `PokeAPIClient.langCodes` now includes `"es"`, since PokéAPI has Spanish species-name coverage.
* No changes under `Sources/PokeTokenBar/UI/` — the Settings language picker iterates `AppLanguage.allCases`, so the new language shows up automatically.
* Targets Spain Spanish (PokéAPI's `es` language code), not `es-419` (Latin American Spanish) — the two differ on a few terms (e.g. `"shiny"` is `variocolor` in `es` vs. `brillante` in `es-419`); this PR is scoped to one language per #149's guidance.

Translation quality disclosure

Per #149's review note: the 25 Pokémon nature names and the 3 item names (Rare Candy, Mint, Shiny Charm) are verified against official PokéAPI Spanish data, not machine-translated. Everything else (~450 remaining UI strings) is machine-translated and has not yet been reviewed by a native speaker.

I put together a side-by-side English → Spanish review page (grouped by section, with search) to make native review easy:
[https://claude.ai/code/artifact/fdfaff9a-6c59-4dc5-8ce6-78332aaa6556](https://claude.ai/code/artifact/fdfaff9a-6c59-4dc5-8ce6-78332aaa6556)

If a native Spanish speaker can review that list, I'll push corrections as a follow-up commit on this PR.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

No changes under Sources/PokeTokenBar/UI/ — adding a language is compiler-guided per #149's architecture and requires no view edits. (The language now appears as an additional option in the existing Settings language picker.)

## Checklist

- [x] `swift build` and `swift test` pass locally
[...]

PR #161, opened 2026-08-14
feat: show version-specific Pokédex entries when a dex cell is tapped

LeeTaek:
## Summary

This PR adds a detail view that shows version-specific Pokédex entries when a Pokémon is tapped in the Collection grid.

The feature helps users who are mainly familiar with earlier generations understand unfamiliar Pokémon. Providing descriptions also makes the Collection feel more like a proper Pokédex rather than only a collection grid.

Previously, tapping a cell also acted as a hidden way to toggle its shiny appearance. Cell taps now open the detail view, while shiny display is controlled by an explicit toggle shared by the grid and the detail view.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## Implementation

Pokédex entries come from PokéAPI's `pokemon-species.flavor_text_entries`.

- A single GraphQL request fetches both the flavor text and the localized version names.
- If GraphQL fails, the request falls back to REST using the existing `speciesCache`.
- The REST fallback may expose version slugs instead of localized names, so its result is not cached. Localized names come back as soon as GraphQL recovers.

The number of available versions differs by language. Korean, Japanese, Spanish and German start at generation VI (up to 12 per species), French starts at Black/White (up to 14), and English goes back to Red/Blue (up to 33). The exact count also varies by species. Versions with no entry in the selected language are omitted rather than filled in from English.

PokéAPI has no `pt` language at all, as #215 already documented for species names. So when the selected language returns no entries at all, the view falls back to English and says so on screen. Those languages already fall back to English for species names, so the screen stays consistent. Partial gaps are not a fallback case, and English itself never falls back.

The detail view replaces the grid and shares its fixed outer height. Entries use a scrollable list rather than pages, because the entry count varies enough by language that paging would turn one screen into a 4-page and an 11-page variant of itself.

## Bug fixes

Two existing issues surfaced while working on this feature, and are fixed here.

1. **Incorrect Japanese API code**

   `ja-Hrkt` now matches the lowercase `ja-hrkt` that PokéAPI actually emits. Since the comparison is a dictionary lookup, the previous spelling never matched. Species names hid this because the trailing `ja` candidate resolved them, but version names have no `ja` entry, so Japanese users would have seen English version titles. The existing test used the same spelling as the code in its fixture, so it never exercised this path.

2. **Incorrect shiny sprite cache state**

   `SpriteLoader.cachedImage(shiny:)` falling back to the regular sprite when no shiny one is cached appears to be intentional. The issue was that `SpriteView.init` recorded the *requested* value as the loaded state. Holding regular pixels while recorded as shiny made `needsReload` return false, so the shiny sprite was never fetched. The seed now reflects the image that was actually returned.

The detail view uses a static PNG sprite. Animated GIFs are tightly cropped while static PNGs use a padded 96×96 canvas, so switching between them makes the sprite visibly change size.

## Merge with `main`

Conflicts with changes that landed after this branch was cut, resolved as follows.

**Representative Pokémon (#158) entry point moved — feedback welcome.** #158 is built on the cell-selection model: selecting a cell surfaced the species info and the pin action in the footer row. Once a tap opens the detail view there is no selection left in the grid to hang that action on, so the pin action moved to the detail header, next to the shiny toggle. It is a species-level action, so the species screen seemed like the natural place. It also frees up the 332pt footer row — `RepresentativeToggleLayoutTests` notes that the species info gets truncated in en/es, and the language count has since grown to six. Cells keep the representative accent background (state), while the selection highlight is gone along with selection itself. `RepresentativeFooterButton` is renamed to `RepresentativeToggleButton` since it no longer lives in a footer. Happy to take a different approach if you'd prefer.

- **Spanish (#159) / French (#185) / Brazilian Portuguese (#215) / German (#234)** — `L.t(...)` now takes seven arguments, so every string added here has `es`, `fr`, `pt` and `de` values. Shiny uses the terms already chosen upstream.
- **`PokeAPIClient.langCodes`** — Those PRs added `es`/`fr`/`pt`/`de` to this line. They are kept as-is, and only `"ja-Hrkt"` is corrected in place to lowercase, so the "union of `apiCodes`" invariant still holds. #234 now enforces that invariant in `PokeAPILanguageTests`, and with `.ja` resolving to `["ja", "ja-hrkt"]` the list is `["ko", "en", "ja", "ja-hrkt", "es", "fr", "pt", "de"]`.
- **Swift 6.3 isolation guard (#216)** — the two new views declare `@MainActor`. `FlavorCacheKey` is split across lines, because the guard matches `": App"` as a substring and a single-line `let language: AppLanguage` trips it.
- `dexRetry` was dropped — `main` already has an equivalent generic `retry`, which the detail view reuses.

## UI changes
스크린샷 2026-08-24 오후 4 26 48

스크린샷 2026-08-24 오후 4 26 54

| Before | After |
| ------ | ----- |
| A cell tap controlled selection and shiny display. | A cell tap opens the version-specific Pokédex entries. |
| Shiny display was a hidden cell-tap behavior. | The grid and the detail view share an explicit `✨ Shiny` toggle. |
| The footer showed species info, the pin action, and pagination. | The footer shows pagination only. |
| Rarity and the pin action lived in the grid footer. | Both live in the detail header. |
| There was no detail view. | Back, sprite, number, rarity, pin action, shiny toggle, and a scrollable entry list. |
| — | Loading, retryable error, English-fallback notice, and no-entry states. |

## Tests

Added `DexFlavorTextTests.swift` covering:

- Flavor-text cleanup that strips game control characters while preserving ideographic spaces
- Localized version names and the slug fallback
- Language selection and release ordering
- The English-fallback decision — empty `pt`/`ko` falls back, empty `en` does not, partial coverage is kept. `pt` is the language that actually takes this path; `de` has 4,718 entries for species 1-649 and takes the normal one
- Language injection, error propagation, and fallback reporting through the store

Also added regression coverage to:

- `ModelLogicTests` for `ja-hrkt` casing and language fallback
- `ShinyCharmTests` for sprite cache seeding

Validation results:

- All 923 tests pass
- `./scripts/test-gate.sh` passes
- Reintroducing either bug makes its regression test fail — verified for both
- `llvm-cov --show-regions` shows no uncovered regions in the new pure-transform branches. The network paths in `PokeAPIClient` stay untested, as that file is excluded from the gate's logic-core set by design

One note on the coverage number the gate prints: it measures a fixed nine-file logic-core set that this PR barely touches, so it is effectively unchanged from the base commit and is not evidence about this change.

## Documentation

Updated `docs/reference/defect-log.md` with three entries:

- Fallbacks hiding an invalid external API identifier
- Recording requested state instead of the actual fallback result
- The size difference between animated and static sprites

## Checklist


chattymin (changes requested):
Please rebase this feature onto the current Pokémon detail screen and add flavor text as a section instead of replacing the existing stats, nature, abilities, moves, and instance selection UI. The request identity must include both species ID and language, and a response should only be applied if the current language still matches. Please add a regression test for out-of-order responses during a language change and keep the existing detail content available on network failure.

PR #162, opened 2026-08-14
Add user-configurable additional scan folders with wildcard support

hannin:
## Problem

Some setups keep Claude session logs outside every root the app scans — for example launchers that isolate each account into its own config directory and inject a per-account `CLAUDE_CONFIG_DIR` only into their child processes. That env var is invisible to both the process environment and the login-shell probe, so the menu bar silently under-reports (on my machine, roughly 3/4 of a day's tokens were missing).

The reader already parses comma-separated `CLAUDE_CONFIG_DIR` values, but GUI users cannot actually use that capability: setting it shell-wide breaks the Claude CLI itself, which reads the value as a single path.

A symlink workaround inside `~/.claude/projects` does not work either — `FileManager.enumerator` does not follow directory symlinks (verified empirically).

## Change

Expose the existing multi-root capability as an app setting instead of hardcoding any specific tool's layout:

- **`customScanRoots` setting (Settings > Advanced)** — comma/newline-separated folders, tilde expansion, per-segment `*` wildcards (via `fnmatch`, so `?`/`[]` come for free). Entries point at log roots directly (no `/projects` suffix assumed), because layouts differ between tools.
- Only existing directories survive expansion, sorted for deterministic root lists. Not-yet-existing patterns are allowed, so instances created later are picked up without touching the setting again.
- Wired into `computeClaudeProjectRoots` — the single source shared by scan/cache/tests. `normalizedRoots` still folds overlaps with the default roots, so a redundant entry costs nothing.
- The settings field shows a live "matches N folders" count, so a typo is visible immediately instead of waiting a refresh cycle. Changing the value invalidates the roots cache and triggers a refresh.
- ko/en/ja strings added.

A whole per-account layout is covered with one wildcard entry (e.g. `<base>/instances/*/projects`); any other custom layout works the same way — no tool-specific code or dependency added.

## Verification

- New tests: glob expansion / existence filtering / ordering (`testCustomScanRootsExpandGlobAndKeepOnlyExistingDirs`), union with defaults and no-op without a value (`testComputeRootsUnionCustomRootsAndDefaultUnchangedWithoutValue`). Full suite: 620 tests, 0 failures.
- Real-machine E2E: with one wildcard entry covering an isolated per-account layout, the menu total matched an independent re-aggregation of all transcripts (same local-day window, keep-max dedup on `(message.id, requestId)`) to the exact token; with the setting empty, behavior is unchanged.

## Notes

- Default behavior is untouched — with the setting empty, the root list is identical to before.
- No new dependencies.

### DIFF
--- a/Sources/PokeTokenBar/Core/LocalUsageReader.swift
+++ b/Sources/PokeTokenBar/Core/LocalUsageReader.swift
@@ -58,6 +58,10 @@ enum LocalUsageReader {
     ///
     /// - `CLAUDE_CONFIG_DIR`: 사용자가 설정 위치를 옮긴 경우. 콤마로 여러 개를 줄 수 있고 각각 `<값>/projects`.
     /// - `~/.config/claude/projects`, `~/.claude/projects`: CLI 기본 위치(전자는 XDG 스타일 설치).
+    /// - 사용자 지정 스캔 폴더(설정): 계정별 격리 래퍼나 커스텀 레이아웃은 세션 로그가
+    ///   위 어디에도 안 남는다. 래퍼가 `CLAUDE_CONFIG_DIR` 를 자식 프로세스에만 주입하면 셸 조회로도
+    ///   못 잡으므로, 사용자가 폴더를 직접 등록하는 것이 유일한 범용 경로다. `*` 와일드카드를 지원해
+    ///   인스턴스가 늘어도 설정을 다시 만질 필요가 없게 한다.
     /// - Claude Desktop 임베디드 세션: 세션 디렉터리마다 CLI 와 같은 모양의 `.claude/projects` 를 갖는다.
     ///   Desktop 으로 일한 사용량이 여기에만 남으므로 빼면 조용히 누락된다.
     /// 계산에 파일시스템 탐색 + (GUI 앱에선) 로그인 셸 조회가 들어가는데 새로고침은 분 단위로 돈다.
@@ -67,6 +71,7 @@ enum LocalUsageReader {
     /// 테스트·진단용 — 캐시를 무시하고 지금 상태로 계산한다.
     static func computeClaudeProjectRoots(
         configDirValue: String? = shellAwareClaudeConfigDir(),
+        customRootsValue: String? = storedCustomScanRoots(),
         home: URL = FileManager.default.homeDirectoryForCurrentUser) -> [URL]
     {
         var roots: [URL] = []
@@ -80,6 +85,9 @@ enum LocalUsageReader {
         }
         roots.append(home.appendingPathComponent(Self.configRelativeProjectsPath))
         roots.append(home.appendingPathComponent(Self.defaultRelativeProjectsPath))
+        if let raw = customRootsValue {
+            roots.append(contentsOf: expandedCustomRoots(raw))
+        }
 
         let desktop = home.appendingPathComponent("Library/Application Support/Claude")
         for store in ["local-agent-mode-sessions", "claude-code-sessions"] {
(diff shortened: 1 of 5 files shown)


chattymin (changes requested):
Thanks for this — and especially for the diagnosis behind it. The `CLAUDE_CONFIG_DIR` capability really was unreachable from a GUI-launched `.app`, the symlink dead-end you documented saved me from suggesting it, and matching the menu total against an independent re-aggregation to the exact token is the kind of verification I wish every PR had. The feature belongs in the app.

I'd like to request some changes before merging. Three are blocking; the rest are recommendations.

## Blocking

### A. Rebase — the branch no longer compiles on `main`

`main` gained a fourth UI language (Spanish) in #159, so `L.t()` is now `t(_ ko:_ en:_ ja:_ es:)` with four required arguments (`Localization.swift:10`). The three new strings here use the old three-argument form.

I resolved the textual conflict locally and built it to confirm:

```
Localization.swift:162: error: missing argument for parameter #4 in call
Localization.swift:166: error: missing argument for parameter #4 in call
Localization.swift:170: error: missing argument for parameter #4 in call
```

CI is green only because it ran against the pre-Spanish merge base. Adding the three `es` strings on top of a rebase was enough — the full suite then passed at 627 tests, 0 failures. So this part is mechanical; nothing is wrong with the logic.

### B. A custom root above the default root silently zeroes all Claude usage

This one is a real defect, and it is reachable by normal typing.

Two existing rules interact badly with a user-supplied root:

1. `normalizedRoots` folds nested roots, so a custom root that is an *ancestor* of `~/.claude/projects` evicts it from the list.
2. `jsonlFiles(in:)` enumerates with `.skipsHiddenFiles`, so the surviving broad root never descends into `.claude`.

Net effect: the total goes to zero, with no error and no log line.

I reproduced it with a scratch test — a temp home containing one assistant entry:

| roots | files found |
| --- | --- |
| default only | 1 |
| `customScanRoots` = the home directory | **0** |

The computed root list collapses to `["<home>"]` and the scan returns nothing.

This is not an exotic input. Because `customScanRoots.didSet` persists and refreshes on every keystroke, typing any `~/...` path passes through the literal value `~`, and that intermediate state is actually scanned and committed. There is a downstream effect too: `CompanionStore.update` rebases a provider's ledger line downward when a value drops (`CompanionStore.swift:377`), so when the real number returns, the day's tokens are credited to the companion a second time. The rebase itself predates this PR; the new setting just makes it easy to trigger.

Suggested fix, smallest version: establish the default roots first and let custom roots only *add* to them, so a custom root can never evict a curated one. Please leave `.skipsHiddenFiles` alone — it is load-bearing elsewhere.

Worth noting that this is the same failure mode #144 set out to remove ("silently under-reports"), arriving through a new door.

### C. Scope the settings key to a provider

The setting is stored under the generic key `customScanRoots` but only feeds `computeClaudeProjectRoots`, so it is Claude-only in effect. Every other provider resolves its root separately, and several of them (OpenCode, Hermes, Copilot, Grok) have the same relocation problem — I'm addressing those separately.

If this ships under a generic key with Claude-only semantics, generalizing later means writing a migration for everyone who set a value in between. Naming it per provider now — `customScanRoots.claude_code` — costs a line today and removes that migration entirely. A plain `String` keyed per provider keeps the `@Observable` binding unchanged, and each reader can later read its own key.

Please also make the UI label Claude-specific. Under the current generic wording ("Additional scan folders") a user will reasonably point it at a Codex or Cursor folder and get nothing back.

## Recommendations (not blocking)

1. **Commit on submit rather than per keystroke.** `TextField` writes the binding continuously, so `didSet` currently fires `UserDefaults` write + roots-cache invalidation + a full `refresh()` on every character. `refresh()` fans out to every provider. `.onSubmit` / focus loss, or a debounce, would be enough — `didSet` can stay as the persistence point.

2. **Move the "matches N folders" count off the main thread.** `expandedCustomRoots` runs `contentsOfDirectory` per wildcard segment and `fileExists` per candidate, synchronously inside `body`. Measured here with the function built standalone: `~/*` 5 ms, `~/*/*` 130 ms, `~/*/*/*` 689 ms. A multi-wildcard pattern will visibly stall the settings window. Computing into `@State` off the main actor would fix it.

3. **Count the roots that survive normalization.** The displayed number is taken before `normalizedRoots` folds duplicates and nesting, so it can claim two folders when only one is scanned. Given the field exists to make a typo visible, the post-fold count is the more honest one.

4. **Keep the existing root tests hermetic.** The new default argument `customRootsValue: String? = storedCustomScanRoots()` means the callers at `LocalUsageReaderTests.swift:184` and `:242` now read `UserDefaults.standard` from whoever runs the suite. It is empty on CI, so it passes — passing `customRootsValue: nil` explicitly at those two call sites keeps them independent of the developer's machine.

5. **Title prefix.** `CONTRIBUTING.md` asks for Conventional Commits, and since the repo squash-merges, the PR title becomes the commit subject on `main` — `feat: add user-configurable additional scan folders with wildcard support`.

Happy to take B myself if you'd rather not carry it — say the word. Otherwise I'll re-review as soon as it's rebased.

chattymin:
Superseded by #187, which generalizes this to every provider. Your `CLAUDE_CONFIG_DIR` diagnosis and the ancestor-root defect are what that PR is built on. Closing here — not a rejection.

PR #163, opened 2026-08-14
fix: stop a second instance from appearing when the login agent is registered

bokeeeey:
## Summary

Enabling "Launch at login" while the app is running leaves two menu bar icons. Registering the LaunchAgent makes launchd start `Contents/MacOS/PokeTokenBar` right away — the plist carries `RunAtLoad` — and LaunchServices' duplicate-launch protection does not apply, because launchd `exec`s the binary directly instead of going through the GUI open path. The same behaviour is reported in [Apple Developer Forums 745712](https://developer.apple.com/forums/thread/745712) (FB13574819), with no DTS reply.

The same thing happens with no user action at all: `migrateFromLegacyLoginItemIfNeeded()` calls `register()` on every launch, so anyone upgrading from the old `SMAppService.mainApp` login item gets a duplicate on first launch.

This adds a startup guard. When another instance of the same bundle started earlier, the newly launched one terminates before it creates its status item, which covers both paths. The exit is clean, so `KeepAlive.SuccessfulExit=false` does not restart it, and from the next login the agent runs alone as designed.

The *later* instance is the one that yields, so the popover or Settings window the user is looking at survives the toggle. When start times are missing or identical the guard deliberately does nothing — a stray extra icon is better than both instances quitting and leaving the user with none.

**Reproducing on `main`**

1. Launch PokeTokenBar — one icon in the menu bar.
2. Open Settings and turn on "Launch at login".
3. A second icon appears immediately, and both instances keep running until one is quit.

스크린샷 2026-08-15 03 22 36

Both icons are the same app running twice — same companion sprite, same token total, and each one opens its own popover. Underneath them are two processes from the same bundle, both parented to launchd, the second starting at the moment the toggle flips:

```
PID    PPID  START     COMMAND
9906   1     00:21:33  /Applications/PokeTokenBar.app/Contents/MacOS/PokeTokenBar
18282  1     00:23:40  /Applications/PokeTokenBar.app/Contents/MacOS/PokeTokenBar

18282   0   io.github.chattymin.poketokenbar.login                     <- the agent
 9906   0   application.io.github.chattymin.poketokenbar.377715942.…   <- the GUI launch
```

Seen on 2.5.1 (Homebrew cask) / macOS 26.5.1 arm64, and present since v2.3.7, the release that introduced the auto-restart login agent.

**Why it was not caught.** `LoginItem` has no tests and is not in `test-gate.sh`'s `LOGIC_CORE`. The manual pass recorded in `c0878c4` exercised `KeepAlive` on an app that had already exited; this defect needs `register()` to run while the app is still up, and the migration path needs an existing `mainApp` login item that a fresh install does not have.

**Changes**

- `Sources/PokeTokenBar/Core/SingleInstance.swift` (new) — `shouldYield(myStartTime:otherStartTimes:)` holds the decision as a pure function, `processStartTime(_:)` reads the kernel's `p_starttime`, and `shouldYieldToRunningInstance()` applies the two to `NSRunningApplication`, gated on `AppEnv.isBundledApp` so `swift test` and raw-binary dev runs never take it.
- `Sources/PokeTokenBar/PokeTokenBarApp.swift` — the guard runs first in `applicationDidFinishLaunching`.
- `Tests/PokeTokenBarTests/SingleInstanceTests.swift` (new) — 11 cases.
- `scripts/e2e.sh` — step 2 now asserts *exactly one* instance instead of "at least one", which is the only automated place the launchd boundary can be checked.
- `docs/reference/defect-log.md` — a new 프로세스·인스턴스 section for the two classes this produced.

**Why not `NSRunningApplication.launchDate`**

That was the first version of this patch, and it silently did nothing. `launchDate` is only recorded for processes LaunchServices started; for one that launchd `exec`'d directly it stays nil indefinitely — and that is precisely the instance which has to yield, so the guard never fired where it mattered. Measured against a live duplicate:

```
pid 9906   p_starttime 15:21:33   launchDate 15:21:33   <- opened normally
pid 18282  p_starttime 15:23:40   launchDate nil        <- started by the login agent
```

Running the decision over those two: on `launchDate` the agent's instance yields `false` and the duplicate survives; on `p_starttime` it yields `true` while the earlier instance stays. The kernel value is present on both paths and comparable across processes.

**Alternatives considered**

The common shape for this in other menu bar apps is to never create the duplicate in the first place: register a small helper as the login item, have it check whether the main app is already running, and either exit or hand off through LaunchServices — [Rectangle](https://github.com/rxhanson/Rectangle/blob/main/RectangleLauncher/AppDelegate.swift) and [Stats](https://github.com/exelban/stats/blob/master/LaunchAtLogin/main.swift) both do exactly that, and Rectangle drops the helper entirely on macOS 13+ in favour of `SMAppService.mainApp`.

Neither fits here, for the same reason: both make the main app start through LaunchServices, which is exactly what takes it out from under launchd — and the agent exists so launchd can act as the crash watchdog (`c0878c4`). Keeping that feature means the app will keep being `exec`'d directly, so it has to recognise its own duplicate.

**The class, swept**

Registering the agent is the only call that starts the binary outside the GUI open path, and it has two sites — `LoginItem.setEnabled(true)` and `migrateFromLegacyLoginItemIfNeeded()` — both covered by the one guard. The other place that restarts the app was checked and left alone: `UpdateChecker.launchDetachedUpgrade` goes through `launchctl kickstart -k` with `open` as its fallback, only after its wait loop has seen the app exit, and `-k` replaces rather than adds. That wait is bounded at 20s, and it is the one spot where this guard could bite — if the wait ever timed out with the app still alive, the kickstarted instance would be the later one, would yield, and the update would not come up.

> **Corrected after review.** This paragraph originally argued that a clean quit is far quicker than the timeout, so the case would not arise. That is wrong, and wrong precisely for the users this PR is for: the loop waits on `pgrep -x PokeTokenBar`, which matches *any* instance, while `terminate` ends only one — so while a duplicate is running the wait always times out. It self-heals and only happens in an already-broken state, so it stays out of this PR; tracked as #175.

The guard sits ahead of `CrashReporter.install` for a second reason besides the icon: after it, the yielding instance would overwrite the running marker and fire `markClean()` on its way out, so a later crash in the surviving instance would read as a clean shutdown on the next launch.

**Why the clean exit is safe**

The app already depends on exactly this. `UpdateChecker.applyUpdate()` schedules a detached `brew upgrade` and then calls `NSApp.terminate(nil)` (`UpdateChecker.swift:65-66`); that flow only works because `KeepAlive.SuccessfulExit=false` does not read a terminate as a crash — otherwise launchd would relaunch the app on top of the running upgrade. The user-facing quit in `PopoverView.swift:598` takes the same path. The guard reuses that exit rather than introducing a new one, so it cannot produce the restart loop a non-zero exit would.

**Two things worth your call**

1. *Which instance survives — and what it costs.* As written the newly launched one yields, so the process that goes is the agent-owned one, and the crash auto-restart shipped in v2.3.7 is effectively off for the rest of that session. It comes back at the next login, and the window is bounded: enabling the toggle is a one-off, and the migration path fires once per upgrade. Still, it is a feature you announced, and this trades it against UI calm.

   The alternative is to keep the agent's process and let the older one go. `launchDate` is the signal for that — it is nil exactly when launchd `exec`'d the process, so the instance under the watchdog is the one reporting nil. That preserves the feature immediately, at the cost of the app disappearing and coming back under the user while they are in Settings. (On the migration path nothing is open, so there the cost is close to zero — a split by open UI is possible too, though it buys a branch.)

   I chose the quieter behaviour, but this weighs a shipped feature against interface calm, which is your call and not mine.
2. *`open -n` for side-by-side inspection.* The guard stops that too. `docs/reference/defect-log.md:266` already records that path as a dead end for energy measurement — the secondary instance does not animate, so it cannot reproduce the 14% idle CPU — and points at replacing the installed primary app instead, so this should not cost an active workflow. If you want an escape hatch anyway it would have to arrive as a launch argument rather than an environment variable, for the reason `defect-log.md:62` already records about GUI-launched apps not inheriting the shell environment.

**On the tests**

The real trigger involves launchd, so it cannot be reproduced in a unit test. The tests lock the two layers that can be: which instance yields (inverse branch, ties, unavailable times) and whether the start time the decision runs on is readable at all — the layer the first version got wrong. Both were checked by injecting the defect rather than trusting a green run: forcing `shouldYield` to `false` fails 3 tests, and forcing `processStartTime` to nil — which reproduces the `launchDate` behaviour — fails 2.

Verified on real builds as well, since the launchd boundary is the part no unit test reaches. Launching the app and then bootstrapping an agent that points at it — `RunAtLoad`, `KeepAlive.SuccessfulExit=false`, `ThrottleInterval 10`, same keys as the shipped plist — leaves **two** instances on `main` and **one** with this branch, with the guard line in the log. The yielding job exits 0 and stays down (`runs = 1`, `last exit code = 0`, rechecked past the throttle interval).

That check now lives in `scripts/e2e.sh`, whose smoke step fails on a second instance rather than passing on "at least one running". That file is your local tool, so say the word and I will drop that hunk.

On coverage, the region report is the number that means anything here. `processStartTime` is 3/3 regions and `shouldYield` fully covered; `shouldYieldToRunningInstance()` is 0/3 because `AppEnv.isBundledApp` is false under `swift test`, which is the same split `AppLog.write` and the limit alerts already use — judgment in a pure function, wiring left uncovered. All six missed regions in the file are in that one function. The gate still reports 89.70%, exactly what `main` reports: the new file is outside `LOGIC_CORE`, so that figure says nothing about this change. Happy to add it to `LOGIC_CORE` if you would rather the guard sat inside the gate.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## Checklist

- [x] `swift build` and `swift test` pass locally — 629 tests, `scripts/test-gate.sh` green


chattymin:
Thanks — the write-up traces the launchd path end to end, and the `launchDate` → `p_starttime` switch is the part I would have gotten wrong.

**Verified locally**

- `size > 0` in `processStartTime` is load-bearing: without it `sysctl` returns rc=0 with size=0 for a dead pid and `p_starttime` reads back as 1970, so the app would yield on every launch. Good catch.
- The guard belongs ahead of `CrashReporter.install` — the `willTerminate` observer is registered inside `install()`, so a yielding instance can't touch the marker.
- Build clean, 629 tests pass, `scripts/test-gate.sh` green at 89.70%.

Worth adding to the case for this: two instances both write `CompanionStore` and `UsageStore` to the same files, so the save is last-writer-wins. This is a data-integrity fix, not a cosmetic one.

**One change before merge**

`AppLog.write` is async and `NSApp.terminate(nil)` reaches `exit(0)` right after, so the guard's log line races teardown — reproducing the sequence in an AppKit app, 42 of 100 runs lost it. That matters because a false positive surfaces as "the app didn't start", which is indistinguishable from a crash without that line. An `AppLog.flush()` (`queue.sync {}`) before `terminate` closes it.

(`CrashReporter.markClean()` has the same shape from the `willTerminate` observer. That one is pre-existing — tracked in #174.)

**Your two calls**

1. Keep the yield direction as written. Unreadable or tied start times leave both instances alive, never zero, and losing the Settings window right after the user flips the toggle reads as a crash. Could you make the gap explicit, though? The comment and the defect-log entry say the watchdog is normal "from the next login" without saying it is off until then, and menu bar users go weeks without logging out.
2. Keep the `scripts/e2e.sh` hunk — it is the only automated place the launchd boundary shows up.

**Follow-up, not this PR**

`launchDetachedUpgrade`'s wait loop is `pgrep -x PokeTokenBar`, which matches any instance, while `terminate` ends only one — so in the duplicate state the 20s wait always times out, and "a clean quit is far quicker than that" does not hold for the population this PR targets. The kickstarted instance then yields and the update does not come up until the next launch. It self-heals and only occurs in an already-broken state, so I would rather not widen this PR. Filed as #175.

Ready to merge otherwise.

bokeeeey:
Both changes are in, plus the framing you suggested.

**`AppLog.flush()` before `terminate`** (b00ee53) — `queue.sync {}` drains the serial queue, so the line lands before `exit(0)`. Verified the way you measured it: a real build, a duplicate raised by an actual LaunchAgent bootstrapped with the shipped plist's keys, 5 of 5 runs keep the line with one instance surviving each time.

No test on `flush()` itself. `write` is gated on `AppEnv.isBundledApp`, so under `swift test` it is a no-op and any assertion I could write would pass with or without the fix. If you see a shape that would actually hold, I will add it.

**The watchdog gap is named now**, in the doc comment and the defect-log entry both. They previously said the watchdog is normal "from the next login", which reads as if nothing is missing before then; they now state that the crash restart is off until then, that menu bar apps go weeks without a logout, and why the other direction was not taken — `launchDate == nil` is what identifies the agent-owned process if you ever want to flip it.

**The data-integrity point reframes this**, and I had missed it entirely. The defect-log entry now leads with it: the icons are the symptom, the last-writer-wins save is the damage (6ba73b7).

**On the update wait loop — you are right, and I was wrong in a way I should have caught.** `pgrep -x` matching any instance is precisely the state this PR is about, so my "a clean quit is far quicker" line fails for exactly the population it was meant to reassure. I corrected that paragraph in the PR body rather than deleting it, so your quote still reads in context, and pointed it at #175.

Thanks for the review — reproducing the log race instead of just flagging it is what made it obvious there was something to fix.

chattymin:

PR #164, opened 2026-08-15
Refactor shared incremental SQLite scan for Cursor and Copilot (#157)

### LINKED ISSUE #157: The Cursor and Copilot readers duplicate the incremental SQLite scan, so a watermark fix has to land twice

shubhransh-gupta:
Fixes #157

## Summary
- Add shared `scanIncrementalSQLiteRoots`, `incrementalScanBoundary`, and `finalizeIncrementalHighWater`
- Route Cursor and Copilot readers through the shared helpers
- Collapse duplicate load-result types via `IncrementalSQLiteLoadResult`

## Test plan
- [ ] `swift test`

### DIFF
--- a/Sources/PokeTokenBar/Core/LocalAdditionalUsageProvider.swift
+++ b/Sources/PokeTokenBar/Core/LocalAdditionalUsageProvider.swift
@@ -334,17 +334,69 @@ enum LocalAdditionalUsageReader {
         } ?? []
     }
 
-    // MARK: Cursor database
+    // MARK: Incremental SQLite scanning (shared)
 
-    struct CursorLoadResult: Sendable {
+    typealias CursorLoadResult = IncrementalSQLiteLoadResult
+    typealias CopilotLoadResult = IncrementalSQLiteLoadResult
+
+    private struct IncrementalDatabaseLoad: Sendable {
+        var entries: [LocalUsageReader.Entry]
+        var highWaterRowID: Int64
+        var didReset: Bool
+    }
+
+    struct IncrementalSQLiteLoadResult: Sendable {
         var entries: [LocalUsageReader.Entry]
         var highWaterByPath: [String: Int64]
-        /// True when any DB was rescanned from scratch (watermark invalidated).
+        /// True when any database was rescanned from scratch (watermark invalidated).
         var didReset: Bool
 
         /// Convenience for single-database tests.
         var highWaterRowID: Int64 { highWaterByPath.values.max() ?? 0 }
     }
+
+    /// Shared shrink detection for incremental SQLite readers (Cursor, Copilot, …).
+    private static func incrementalScanBoundary(
+        afterRowID: Int64,
+        maxRowID: Int64?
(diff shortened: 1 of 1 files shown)


shubhransh-gupta:
Hi — gentle ping when you have a moment 👋

CI is green (`build-test`, `secret-scan`). This is scoped to the shared SQLite scan refactor from #157 — happy to adjust anything if the approach doesn't match how you want the Copilot/Cursor watermark logic structured.

Thanks for maintaining PokeTokenBar!

chattymin:
Thanks for your interest in pokeTokenBar and for taking the time to contribute!
However, since [another PR](https://github.com/chattymin/PokeTokenBar/pull/160) with the same changes was already in progress, we've decided to close this PR.

I really appreciate your contribution!

shubhransh-gupta:
No worries at all — thanks for the quick heads-up and for merging #160! Glad the refactor landed; I'll look for another good first issue in PokeTokenBar. Appreciate you maintaining the project.

Issue #168, opened 2026-08-15
Make the popover's outside-click monitor registration idempotent


chattymin:
## What

`startOutsideClickMonitor()` (`Sources/PokeTokenBar/PokeTokenBarApp.swift:375`) overwrites
`outsideClickMonitor` without removing whatever token is already stored:

```swift
private func startOutsideClickMonitor() {
    outsideClickMonitor = NSEvent.addGlobalMonitorForEvents(matching: [.leftMouseDown, .rightMouseDown]) { ... }
}
```

Start and stop are not symmetric either: registration happens at the call site
(`togglePopover()`, `PokeTokenBarApp.swift:360`) while removal happens in the delegate callback
(`popoverDidClose`, `:368`).

## Why it matters

Nothing is broken today. `togglePopover()` is the only path that calls `popover.show`, and
`openPopover()` (`:336`) guards on `popover.isShown`, so the monitor cannot be registered twice.

The point is what happens when a second show path is added: the previous token is dropped on the
floor, `NSEvent.removeMonitor` is never called for it, and a global mouse monitor stays installed for
the rest of the process — outliving the popover it was meant to track. That failure is silent; there is
no crash and no visible symptom, which is exactly the kind of thing worth locking down mechanically
while the code is small.

## Suggested fix

Either of these is enough — the second is the tidier one since it makes the pairing structural:

1. Make registration idempotent — call `stopOutsideClickMonitor()` first, or `guard outsideClickMonitor == nil`.
2. Move registration into the `popoverDidShow(_:)` delegate callback so start and stop are both
   delegate-driven, mirroring `popoverDidClose`.

If option 2 is taken, check that a click landing during the show animation still dismisses as expected.

## Notes

This came out of the review on #154 and was deliberately left out of that PR to keep it scoped.

shubhransh-gupta:
Opened a PR for this — happy to close in favor of #172 if that lands first (same approach: `OutsideClickMonitor` + `popoverDidShow` + idempotent start).

PR #169, opened 2026-08-15
docs: suggest idempotent popover registration (refs #168)

MaxCode917:
Add a short note documenting a recommended idempotent registration pattern for the popover outside-click monitor. Helps maintainers reproduce and patch the issue (refs #168).

### DIFF
--- a/docs/notes/popover-idempotent.md
+++ b/docs/notes/popover-idempotent.md
@@ -1,0 +1,3 @@
+# Popover idempotent registration
+
+This documents a suggested change to make the popover outside-click registration idempotent to avoid duplicate handlers.


chattymin:
Thanks for looking at #168, @MaxCode917. Closing this one — it doesn't move the issue forward.

1. **No code change.** `startOutsideClickMonitor()` (`PokeTokenBarApp.swift:375`) still overwrites `outsideClickMonitor` without removing the stored token.
2. **The note says less than the issue.** #168 already has the repro condition and two concrete fix options.
3. **`docs/notes/` isn't this repo's docs layout.** Reference docs live in `docs/reference/` with frontmatter and a `CLAUDE.md` index entry.

If you want to take #168: add `guard outsideClickMonitor == nil else { return }`, or move registration into `popoverDidShow(_:)` (then manually check a click during the show animation still dismisses). Please fill out the PR template. Happy to review.

Issue #167, opened 2026-08-15
Speech-bubble layout guard passes copy that the view truncates


chattymin:
## What

`testLocalizedAlertBubbleFitsDefaultPanel` (`Tests/PokeTokenBarTests/PerformanceTests.swift:204`) is the
guard that keeps alert copy from being clipped in the floating pet's speech bubble — it exists because
Japanese copy once clipped by ~9pt (#124). It measures with
`FloatingPetController.measureSpeechBubble` (`Sources/PokeTokenBar/UI/FloatingPetPanel.swift:112`),
which lays the text out with **no height constraint** (`height: 10_000`) and then asserts the result fits
`bubbleHeadroom` (72pt, minus 2).

The view it is guarding renders with `.lineLimit(2)` (`FloatingPetPanel.swift:417`).

So the guard answers "does the bubble overflow the panel?" while the view's actual failure mode is
"does the copy exceed two lines and get truncated?" Those are different thresholds, and the gap is one
line wide.

## Evidence

Reproducing `measureSpeechBubble`'s math on a 164pt content column (`bubbleMinWidth` 180 −
`bubbleHorizontalPadding` 8 × 2), title `Límite inminente` at 11pt bold, body at 10pt:

| body | measured height | guard limit (70pt) | rendered with `lineLimit(2)` |
|---|---|---|---|
| current es — `Sesión de 5 horas de Claude al 85%` (2 lines) | 57pt | passes | fine |
| longer copy that wraps to 3 lines | 69pt | **passes** | **truncated** |
| copy that wraps to 4 lines | 81pt | fails | truncated |

Confirmed by injecting the defect rather than trusting the green run: replacing the Spanish 5-hour-session
label with progressively longer copy leaves the guard green through three lines and only fails at four.

Separately, the width assertion cannot fail at all: `measureSpeechBubble` returns
`min(contentWidth, max(titleWidth, bodyWidth)) + padding`, which is `panel.width` by construction.

None of the four shipped languages currently exceeds two lines, so nothing is clipped today. The problem
is that the guard will stay green for the copy that does.

## Suggested fix

Measure the way the view draws:

- Cap the body layout at two lines (or have `measureSpeechBubble` report the line count) and fail when
  the copy would be truncated, not only when the panel would overflow.
- Drop or rewrite the width assertion so it can actually fail — e.g. compare the *unclamped* text width
  against the content column.

## How to verify

The guard should fail when copy is too long, so check it by injection rather than by a green run:
[...]

PR #171, opened 2026-08-16
fix: reject speech-bubble copy that the view would truncate (#167)

Borisserz:
## Summary

`testLocalizedAlertBubbleFitsDefaultPanel` was added after Japanese copy clipped the bubble by ~9pt (#124). It measured with `measureSpeechBubble`, which lays text out at `height: 10_000`, then asserted the chrome fits `bubbleHeadroom` (72pt − 2).

`SpeechBubbleView` draws the body with `.lineLimit(2)`. Those are different thresholds. Owner's table on #167: 3-line copy measures ~69pt (guard passes) and is truncated in the view. 4-line copy is the first height that fails (~81pt). None of the four shipped languages wrap past two lines today, so the guard stayed green for the copy that would actually clip.

The width assert could never fail either: `measureSpeechBubble` returns `min(contentWidth, …) + padding`, which is `panel.width` by construction.

`measureSpeechBubbleLayout` now reports the signals the view fails on:

- `bodyLineCount` / `wouldTruncate` — wrap count at the 164pt column vs `bubbleBodyLineLimit` (shared with the view; a literal `2` in only one place can drift)
- `unclampedTitleWidth` / `unclampedBodyWidth` — single-line width, not `min(contentWidth, …)`
- `size` stays unconstrained so the 4-line overflow assert can still go red

No visible bubble change. `Localization.swift` is untouched.

## Type of change

- [x] Bug fix

## UI changes

No visible change. `SpeechBubbleView` now reads `.lineLimit(FloatingPetController.bubbleBodyLineLimit)` instead of a literal `2` — same value, so measure and view cannot drift.

## Checklist

## Tests

`FloatingPetEnergyTests`:

- `testThreeLineBodyThatFitsHeadroomWouldTruncate` — 3-line body, height ≤ 70pt, `wouldTruncate == true`. This is the #167 gap: a headroom-only guard stays green here.
- `testTwoLineBodyDoesNotTruncate` — two-line wrap is the view's designed capacity.
- `testFourLineBodyExceedsHeadroomAndWouldTruncate` — the other row of the table (truncate **and** overflow). A∥B: truncate-only and overflow are separate tests.
- `testUnclampedBubbleTextWidthCanExceedContentColumn` — `unclampedBodyWidth > size.width`, so the width check is not `min(contentWidth, …) + padding`.
- `testLocalizedAlertBubbleFitsDefaultPanel` — every `AppLanguage`, both titles, every window `buildLimitWindows` can put in a body. Fails on `wouldTruncate`, not on clamped width.

The 2/3/4-line fixtures grow a wrapping string until unconstrained `measureSpeechBubble` height jumps, so they do not hardcode a twin of `bodyLineCount`.

Confirmed by injection, not by a green run: lengthening the Spanish `claudeFiveHour` until the body wrapped to 3 lines made `testLocalizedAlertBubbleFitsDefaultPanel` fail (`es` / `Límite inminente` / 3 lines). Restored afterwards. `Localization.swift` is not in this diff.

`swift test --filter FloatingPetEnergyTests`: 12 tests, 0 failures.

Full suite: 629 pass, 9 skipped, 0 failures.

## How to verify

```bash
  swift test --filter FloatingPetEnergyTests
```

To re-check the injection: temporarily lengthen one `claudeFiveHour` string in `Localization.swift` until it wraps to three lines, run `--filter testLocalizedAlertBubbleFitsDefaultPanel` (must go red), then restore.

### DIFF
--- a/Sources/PokeTokenBar/UI/FloatingPetPanel.swift
+++ b/Sources/PokeTokenBar/UI/FloatingPetPanel.swift
@@ -32,6 +32,19 @@ final class FloatingPetController: NSObject, NSWindowDelegate {
     static let bubbleHorizontalPadding: CGFloat = 8
     /// Fixed text column — wraps instead of growing past the panel (`bubbleMinWidth` − 16).
     static let bubbleContentWidth: CGFloat = bubbleMinWidth - (bubbleHorizontalPadding * 2)
+    /// `SpeechBubbleView` body `.lineLimit`. Measure and view must share this — a
+    /// headroom-only guard stays green for 3-line copy that still fits 70pt (#167).
+    static let bubbleBodyLineLimit = 2
+
+    /// Chrome size plus the signals the view actually fails on: wrap count vs
+    /// `bubbleBodyLineLimit`, and single-line width vs the content column.
+    struct SpeechBubbleLayout: Equatable {
+        var size: NSSize
+        var bodyLineCount: Int
+        var unclampedTitleWidth: CGFloat
+        var unclampedBodyWidth: CGFloat
+        var wouldTruncate: Bool
+    }
 
     private var onOpenPopover: (() -> Void)?
     private var onHide: (() -> Void)?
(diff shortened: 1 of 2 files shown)


chattymin:

PR #172, opened 2026-08-16
fix: make the popover outside-click monitor idempotent (#168)

Borisserz:
## Summary

`startOutsideClickMonitor()` assigned a new `NSEvent` global-monitor token over the previous one. `togglePopover()` registered; `popoverDidClose` removed. Nothing leaks today — `openPopover()` guards `isShown`, and `show` has one call site — but a second show path would drop the first token, never call `removeMonitor`, and leave a process-lifetime observer with no crash and no visible symptom. Leftover from the #154 review; #169 was docs-only and correctly closed.

Both options from #168:

1. **Registration moved to `popoverDidShow`** so start and stop are both delegate-driven.
2. **`start` is idempotent** (`guard token == nil`) so a double `didShow` cannot overwrite a live token.

Register/remove are injected on a small `OutsideClickMonitor` so the leak is testable without installing a real monitor. Dropping the `guard` makes `testStartTwiceKeepsFirstTokenAndDoesNotRegisterAgain` fail: two registers, `stop` removes only the second token.

No docs. `docs/notes/` is not this repo's layout (#169).

## Type of change

- [x] Bug fix

## UI changes

No visible change. The popover still closes on an outside click; only the monitor's lifetime pairing changed.

## Checklist

## Tests

`OutsideClickMonitorTests` — each branch alone:

- `testStartTwiceKeepsFirstTokenAndDoesNotRegisterAgain` — the #168 leak. Second start must not register; `stop` must remove token `1`, not a replacement.
- `testStopWhenIdleDoesNotRemove` — `popoverDidClose` after a show that never registered.
- `testStartAfterStopRegistersAgain` — open → close → open.
- `testNilRegisterDoesNotOccupyTheSlot` — `addGlobalMonitorForEvents` is `Optional`; a nil result must not block the next start.

Confirmed red-then-green: the first implementation assigned `token = register()` (today's production bug); the twice-start test failed with `registers == 2` and `removed == [2]`. The `guard` flipped it.

`swift test --filter OutsideClickMonitorTests`: 4 tests, 0 failures.

Full suite: 629 pass, 9 skipped, 0 failures.

## How to verify

```bash
  swift test --filter OutsideClickMonitorTests
```

Manual (owner note on #168 — a click during the show animation):

1. Click the menu-bar icon. Popover opens.
2. Click the desktop. Popover closes.
3. Click the icon, then immediately click the desktop while it is still appearing. It must still dismiss (`.transient` handles that frame; `didShow` then installs the monitor for later outside clicks).
4. Open the popover, then open Control Center or another menu-bar extra. The popover must close — that is the path `.transient` misses, which is why the monitor exists.
5. Repeat 1–4 a few times. No second icon, no leftover click-to-dismiss after the popover is gone.

### DIFF
--- a/Sources/PokeTokenBar/OutsideClickMonitor.swift
+++ b/Sources/PokeTokenBar/OutsideClickMonitor.swift
@@ -1,0 +1,22 @@
+import Foundation
+
+/// Token holder for the popover's `NSEvent` global mouse monitor (#168 / leftover of #154).
+///
+/// Start and stop are paired: a second `start` must not overwrite a live token, or
+/// `removeMonitor` never sees it and the observer outlives the popover. Register and
+/// remove are injected so tests can prove the leak without installing a real monitor.
+struct OutsideClickMonitor {
+    private(set) var token: Any?
+
+    mutating func start(register: () -> Any?) {
+        guard token == nil else { return }
+        token = register()
+    }
+
+    mutating func stop(remove: (Any) -> Void) {
+        if let token {
+            remove(token)
+            self.token = nil
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


chattymin:

PR #173, opened 2026-08-16
fix: make the popover outside-click monitor idempotent (#168)

### LINKED ISSUE #168: Make the popover's outside-click monitor registration idempotent

shubhransh-gupta:
## Summary

`startOutsideClickMonitor()` could overwrite a live `NSEvent` token without calling `removeMonitor`, leaking a process-lifetime global observer if a second show path is added (#168).

- Extract `OutsideClickMonitor` with idempotent `start` / paired `stop`
- Register in `popoverDidShow` (delegate-driven, mirroring `popoverDidClose`)
- Remove registration from `togglePopover()` show path

## Type of change

- [x] Bug fix

## UI changes

No visible change — outside clicks still dismiss the popover; only monitor lifetime pairing changed.

## Checklist

- [x] Tests added for the idempotent registration contract

## Tests

`OutsideClickMonitorTests` (4 cases) — inject register/remove so the leak is testable without a real global monitor:

- Second `start` is a no-op and `stop` removes the first token
- `stop` when idle does not call remove
- `start` after `stop` registers again
- Nil register result does not block a subsequent start

```
  swift test --filter OutsideClickMonitorTests
```

Fixes #168

### DIFF
--- a/Sources/PokeTokenBar/OutsideClickMonitor.swift
+++ b/Sources/PokeTokenBar/OutsideClickMonitor.swift
@@ -1,0 +1,22 @@
+import Foundation
+
+/// Token holder for the popover's `NSEvent` global mouse monitor (#168 / leftover of #154).
+///
+/// Start and stop are paired: a second `start` must not overwrite a live token, or
+/// `removeMonitor` never sees it and the observer outlives the popover. Register and
+/// remove are injected so tests can prove the leak without installing a real monitor.
+struct OutsideClickMonitor {
+    private(set) var token: Any?
+
+    mutating func start(register: () -> Any?) {
+        guard token == nil else { return }
+        token = register()
+    }
+
+    mutating func stop(remove: (Any) -> Void) {
+        if let token {
+            remove(token)
+            self.token = nil
+        }
+    }
+}
--- a/Sources/PokeTokenBar/PokeTokenBarApp.swift
+++ b/Sources/PokeTokenBar/PokeTokenBarApp.swift
@@ -17,7 +17,7 @@ struct PokeTokenBarApp: App {
 final class AppDelegate: NSObject, NSApplicationDelegate, NSPopoverDelegate {
     private var statusItem: NSStatusItem!
     private let popover = NSPopover()
-    private var outsideClickMonitor: Any?
+    private var outsideClickMonitor = OutsideClickMonitor()
     private var store: UsageStore!
     private var companion: CompanionStore!
     private var updater: UpdateChecker!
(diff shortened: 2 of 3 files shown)


shubhransh-gupta:
Hi — review ping when convenient 👋

Fixes #168: idempotent outside-click monitor via `OutsideClickMonitor` + `popoverDidShow` registration. `OutsideClickMonitorTests` (4 cases) pass locally; full suite not re-run but change is isolated.

Note: #172 covers the same issue with a parallel approach — happy to close this one if you prefer that PR. Thanks!

chattymin:
Thanks for taking this on, @shubhransh-gupta — and for the note on #168 offering to close if #172 landed first.

#172 was opened first and has now been merged, so I'm closing this one. Both PRs took the same approach and the fix is equivalent.

Sorry this is the second duplicate you've hit here after #164 — that's timing, not anything on your end.

I really appreciate the contributions.

Issue #174, opened 2026-08-16
AppLog drops lines written on the way out of the process, so `clean shutdown` is usually lost


chattymin:
`AppLog.write` hands the line to a background queue (`AppLog.swift:28`) and returns. When the caller is on its way out of the process, the line races the teardown and is frequently lost.

The site that matters today is `CrashReporter.markClean()` (`CrashReporter.swift:65-68`), which runs from the `NSApplication.willTerminateNotification` observer registered in `install()` (`:45-47`). AppKit posts that notification and then exits, so `clean shutdown` is written only if the log queue happens to get scheduled first.

Measured with an AppKit app reproducing the sequence — enqueue on a serial `DispatchQueue`, then `NSApp.terminate(nil)`:

```
runs=100 landed=58 lost=42
```

Consistent with what is on disk here: the current 1.2M `PokeTokenBar.log` has 1 `launch:` line and 0 `clean shutdown` lines; the rotated 2.0M `PokeTokenBar.old.log` has 1 and 1.

The consequence is not just a missing line. `clean shutdown` is the human-readable half of the shutdown record, so the log reads as if sessions end abruptly even when they don't. The marker file itself is removed synchronously (`FileManager.removeItem`, `:66`), so the *detection* in `install()` `:37` is unaffected — this is a diagnostics defect, not a false crash report.

Fix shape: an `AppLog.flush()` that does `queue.sync {}`, called on the paths that exit right after logging. Sweep for callers, not just this one — #163 adds a second such site in `applicationDidFinishLaunching`.

Class, for `docs/reference/defect-log.md`: **an async logger is not a logger on an exit path.** Any `AppLog.write` whose caller reaches `exit()` in the same turn needs a flush, and the absence of the line cannot be used as evidence that the branch did not run.

PR #180, opened 2026-08-16
fix: flush AppLog on the willTerminate exit path (#174)

Borisserz:
## Summary

`AppLog.write` enqueues on a background queue and returns. `CrashReporter.markClean()` runs from the `willTerminate` observer; AppKit then exits, so `clean shutdown` is written only if the log queue happens to get scheduled first. Owner measured 42 of 100 lost. The marker file is removed synchronously, so crash *detection* is fine — the log just reads as if every session ended abruptly.

#163 already added `AppLog.flush()` and used it on the duplicate-yield path. This PR is the remaining site from that review, plus the sweep: both write-then-exit callers now go through `writeAndFlush`.

`write` is gated on `AppEnv.isBundledApp`, so a test that opened the production log would pass with or without the fix (the gap #163 left). The drain is pinned on `AsyncLineLog` with an injected sink — the same injection shape as #168. Dropping `flush` from `writeAndFlush` makes `testWriteAndFlushLandsTheLineBeforeReturning` fail (`landed == []`).

Class recorded in `docs/reference/defect-log.md`: an async logger is not a logger on an exit path.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## UI changes

No visible change. Only the shutdown / yield log line is now guaranteed to land.

## Checklist

## Tests

`AppLogTests` — each branch alone:

- `testFlushDrainsAQueuedWriteBeforeReturning` — the #174 leak. `write` returns before the sink runs; `flush` waits. A 30ms sink makes the race visible without sleeping the test thread after flush.
- `testFlushDrainsEveryWriteQueuedBeforeIt` — `queue.sync {}` is a barrier behind every earlier `async`, not only the last line.
- `testWriteAndFlushLandsTheLineBeforeReturning` — the call-site pair. Confirmed red-then-green: removing `flush` from `writeAndFlush` fails with `landed == []`.
- `testMarkCleanUsesWriteAndFlushNotWriteAlone` — source scan of `markClean`. `AppLog.write` is a no-op under `swift test`, so calling `markClean` cannot prove the drain (same split as `testNoProviderReadsUsageLocationEnvDirectly`).
- `testDuplicateYieldUsesWriteAndFlushBeforeTerminate` — the #163 site, swept onto the same pair so a future edit cannot drop one and keep the other.

`swift test --filter AppLogTests`: 5 tests, 0 failures.

Full suite: 681 pass, 9 skipped, 0 failures.

## How to verify

```bash
  swift test --filter AppLogTests
```

Manual: quit the bundled app from the popover. `~/Library/Logs/PokeTokenBar.log` should contain a `clean shutdown` line for that session, not only `launch:`.

### DIFF
--- a/Sources/PokeTokenBar/Core/AppLog.swift
+++ b/Sources/PokeTokenBar/Core/AppLog.swift
@@ -43,10 +43,11 @@ enum AppLog {
     }
 
     /// Exit-path pair: enqueue, then wait until the line has been sunk.
-    /// Callers that reach `exit()` after `write` alone lose the line (#174).
+    /// Routes through the tested `backend.writeAndFlush` — a second
+    /// `write()`+`flush()` pair is untested and reintroduces #174.
     static func writeAndFlush(_ message: String) {
-        write(message)
-        flush()
+        guard AppEnv.isBundledApp else { return }
+        backend.writeAndFlush(formatted(message))
     }
 
     static func write(_ message: String) {
@@ -54,8 +55,11 @@ enum AppLog {
         // 않게(형제 write 경로 writeParitySnapshot·checkLimitNotifications 와 동일 가드). 테스트가
         // 크래시 진단 로그에 fixture 값을 남기고 회전으로 실이력을 밀어내던 결함 차단.
         guard AppEnv.isBundledApp else { return }
-        let line = "[\(ISO8601DateFormatter().string(from: Date()))] \(message)\n"
-        backend.write(line)
+        backend.write(formatted(message))
+    }
+
+    private static func formatted(_ message: String) -> String {
+        "[\(ISO8601DateFormatter().string(from: Date()))] \(message)\n"
     }
 }
 
--- a/Sources/PokeTokenBar/Core/AppLog.swift
+++ b/Sources/PokeTokenBar/Core/AppLog.swift
@@ -12,18 +12,41 @@ enum AppLog {
         return dir.appendingPathComponent("PokeTokenBar.log")
     }()
 
-    private static let queue = DispatchQueue(label: "poketokenbar.log")
-
     /// 로그 상한 — 초과 시 .old 로 1세대 회전(무한 증가 방지, 디스크 상한 ≈ 2×maxBytes = 4MB).
     /// 24/7 메뉴바 앱이라 회전이 잦아 진단 이력이 금방 사라지던 것 완화(장애 직전 컨텍스트 보존).
     /// 레퍼런스(size-capped 회전, 수 MB)에 맞춤 — 일자별 폴더는 무한 성장/정리 필요라 채택 안 함.
     private static let maxBytes = 2 * 1024 * 1024
 
+    private static let backend = AsyncLineLog(queue: DispatchQueue(label: "poketokenbar.log")) { line in
+        if let size = try? FileManager.default.attributesOfItem(atPath: url.path)[.size] as? Int,
+           size > maxBytes {
+            let old = url.deletingPathExtension().appendingPathExtension("old.log")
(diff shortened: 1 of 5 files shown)


chattymin:
Thanks @Borisserz — the sweep is complete and the defect-log entry captures the class well. I checked every `NSApp.terminate` / `exit(` site; the other two (`UpdateChecker.swift:66`, `PopoverView.swift:598`) have no preceding `write`, so this PR covers the write-then-exit set.

Requesting one change before merge, plus one I'd like your read on.

### 1. The regression test doesn't pin the production `writeAndFlush` — blocking

I removed `flush()` from `AppLog.writeAndFlush`, which reintroduces #174 exactly, and the full suite stayed green:

```swift
static func writeAndFlush(_ message: String) {
    write(message)
    // flush()   ← removed
}
```

```
swift test → Executed 681 tests, with 7 tests skipped and 0 failures
```

There are two `writeAndFlush` implementations. `AsyncLineLog.writeAndFlush` is the one `testWriteAndFlushLandsTheLineBeforeReturning` covers, and it has no production caller (`rg backend.writeAndFlush` → 0). `AppLog.writeAndFlush` is the one `markClean` and the yield path call, and nothing pins it — the two source scans assert that the *call site* spells `writeAndFlush`, not that `writeAndFlush` drains.

The red-then-green claim in the description holds for the test-only variant, not the shipped one.

Routing `AppLog.writeAndFlush` through `backend.writeAndFlush` would leave one tested implementation and retire the unused one. One caution if you take that shape: keep the `AppEnv.isBundledApp` guard on the path. Bypassing `write()` without it lets `swift test` write to the real log, which is what that guard was added to prevent.

### 2. `functionBody` over-captures and rejects correct code — non-blocking

In `testDuplicateYieldUsesWriteAndFlushBeforeTerminate`, the helper returns lines 36–347 of `PokeTokenBarApp.swift` for `applicationDidFinishLaunching` — 311 lines spanning about ten methods. The `??` prefers the next `\n    func ` (line 347) over the next `\n    ///` (line 88).

`XCTAssertFalse(body.contains("AppLog.write("))` then fires on any write anywhere in that span. Adding a correct one that is not on an exit path:

```swift
signal(SIGPIPE, SIG_IGN)
AppLog.write("launch: SIGPIPE ignored")
```

```
AppLogTests.swift:86: XCTAssertFalse failed - write() without flush on this path is the race #163 already closed
```

The message names a race that doesn't apply to that line. Narrowing the scan to the `shouldYieldToRunningInstance()` block, or keeping only the positive assertion, would both close it — your call which.

The `markClean` scan is fine, for what it's worth: `CrashReporter` is all `static func`, so the `func` marker never matches and it stops at the next doc comment.

### 3. Note

`AsyncLineLog.writeAndFlush` has no production caller today. Fixing 1 likely removes it.

Borisserz:
Addressed both notes.

**1.** `AppLog.writeAndFlush` now calls `backend.writeAndFlush` (the pair `testWriteAndFlushLandsTheLineBeforeReturning` actually runs) and keeps the `AppEnv.isBundledApp` guard so `swift test` still cannot touch the production log. `testAppLogWriteAndFlushRoutesThroughBackend` pins that shape: dropping `flush()` from a second `write()`+`flush()` pair no longer stays green.

**2.** The yield scan is the `shouldYieldToRunningInstance()` block only, up to `signal(SIGPIPE`. A launch-path `AppLog.write` after that is not an exit-path race and no longer fails the test.

`swift test --filter AppLogTests`: 6 tests, 0 failures. Full suite: 682 pass, 9 skipped, 0 failures.

chattymin:

PR #181, opened 2026-08-17
fix: preserve Codex usage after session archiving

ite476:
## Summary

- Scan both active and archived Codex rollout roots.
- Preserve usage when Codex moves a rollout from `sessions` to `archived_sessions`.
- Deduplicate the same rollout when it is temporarily visible in both roots.
- Make the default Codex root composition injectable so tests exercise the same source used by the reader and cache.
- Normalize Codex scan roots before enumerating them, including symlinked and nested roots.
- Document the Codex active/archive root rule in `docs/reference/provider-extension.md`.
- Add regression tests for default root composition, archive relocation, cross-root duplication, and symlinked roots.

## Root cause

Codex defines both `sessions` and `archived_sessions` as local rollout locations. Its archive implementation moves rollout files with `std::fs::rename()` and then updates the archived thread metadata:

- [Codex rollout roots](https://github.com/openai/codex/blob/2bd8727a0c07655fafeae0e4b6fe4c2b4741b399/codex-rs/rollout/src/lib.rs#L67-L68)
- [Codex archive implementation](https://github.com/openai/codex/blob/fa9a05f2d25f95b39cf365e77cb10f87e8719185/codex-rs/thread-store/src/local/archive_thread.rs#L76-L115)
- [Codex archive relocation test](https://github.com/openai/codex/blob/fa9a05f2d25f95b39cf365e77cb10f87e8719185/codex-rs/thread-store/src/local/archive_thread.rs#L260-L279)

PokeTokenBar previously scanned only `~/.codex/sessions`, so an archived rollout appeared to disappear from the daily total. The two directories are alternative locations for the same local rollout history, not independent usage streams.

Codex records session identity and cumulative token usage in the rollout format. The existing resolver already uses stable session/state-based canonical IDs, so this change reuses that deduplication boundary instead of summing both directory trees blindly. Codex documents `total_token_usage` as accumulated session usage and `last_token_usage` as the latest active context usage in its TUI model:

- [Codex token usage model](https://github.com/openai/codex/blob/f86d95a242648f91c2bb2b55ed336d9e4d95762e/codex-rs/tui/src/token_usage.rs#L37-L40)

## Scope

This fixes usage loss caused by normal `sessions` → `archived_sessions` relocation and prevents duplicate counting during a temporary cross-root overlap.

This PR does not attempt to solve torn reads during an active write, corrupted or missing rollout files, custom Codex homes, account switching, or server-side quota attribution.

## Validation

- `swift test --filter LocalUsageReaderTests` — 56 tests passed.
- `./scripts/test-gate.sh` — 680 tests passed, 7 skipped, 0 failures.
- Logic-core line coverage: 89.67% (required: 75%).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## Checklist

- [x] `swift build` and `swift test` pass locally.
- [x] PR title and description are written in English.
- [x] No UI changes.
- [x] No copyrighted assets, secrets, or private tooling references are included.
- [x] Tests were added or updated for this change.

### DIFF
--- a/Sources/PokeTokenBar/Core/LocalUsageCache.swift
+++ b/Sources/PokeTokenBar/Core/LocalUsageCache.swift
@@ -86,6 +86,7 @@ actor LocalUsageCache {
     // 테스트 시임 — 기본값은 실환경(실 로그 루트·Application Support·실시간).
     private let claudeRoot: URL?
     private let codexRoot: URL?
+    private let codexRoots: [URL]?
     private let geminiRoot: URL?
     private let grokRoot: URL?
     private let fileURL: URL
@@ -98,7 +99,8 @@ actor LocalUsageCache {
     private let claudeRoots: [URL]?
 
     init(claudeRoot: URL? = nil, claudeRoots: [URL]? = nil,
-         codexRoot: URL? = nil, geminiRoot: URL? = nil, grokRoot: URL? = nil,
+         codexRoot: URL? = nil, codexRoots: [URL]? = nil,
+         geminiRoot: URL? = nil, grokRoot: URL? = nil,
          fileURL: URL? = nil, now: @escaping @Sendable () -> Date = Date.init,
          codexProbe: @escaping @Sendable (URL) throws -> String? = {
              try LocalUsageReader.probeCodexRolloutSessionID(at: $0)
@@ -109,6 +111,7 @@ actor LocalUsageCache {
         self.claudeRoots = claudeRoots
         self.claudeRoot = claudeRoot
         self.codexRoot = codexRoot
+        self.codexRoots = codexRoots
         self.geminiRoot = geminiRoot
         self.grokRoot = grokRoot
         self.fileURL = fileURL ?? Self.defaultFileURL
@@ -143,8 +146,9 @@ actor LocalUsageCache {
     func codexEntries(modifiedSince: Date) -> [LocalUsageReader.Entry] {
         ensureLoaded()
         let fmt = LocalUsageReader.localDayFormatter()
+        let roots = codexRoots ?? codexRoot.map { [$0] } ?? LocalUsageReader.codexScanRoots
         let (rollouts, includedPaths) = collectCodexRollouts(
-            root: codexRoot ?? LocalUsageReader.codexSessionsDir,
+            roots: roots,
             since: modifiedSince,
             fmt: fmt
         )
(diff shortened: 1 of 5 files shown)


chattymin:
Thanks @ite476 — the root cause holds up. I pulled the Codex sources rather than take the description's word for it: `ARCHIVED_SESSIONS_SUBDIR` is defined in `rollout/src/lib.rs`, the archive path goes through `std::fs::rename`, and upstream has a test named `archive_thread_moves_rollout_to_archived_collection`. The move semantics are what you describe.

Requesting one change before merge; the other two are notes.

### 1. The tests don't share the single root source — blocking

I reverted the user-visible half of the fix and the full suite stayed green:

```swift
static var codexScanRoots: [URL] {
    [codexSessionsDir]        // archived root dropped
}
```

```
swift test → Executed 678 tests, with 7 tests skipped and 0 failures
```

Both new tests inject `codexRoots: [root, archivedRoot]`, so nothing exercises `codexScanRoots`. `docs/reference/provider-extension.md` asks for the opposite — the scanner, the cache, and the tests should share that one list.

There is a seam for this already: `claudeProjectRoots` has `computeClaudeProjectRoots(configDirValue:home:)`, and `LocalUsageReaderTests` pins the composition through it with a fake home. `codexScanRoots` reads `homeDirectoryForCurrentUser` directly, so it can't be pinned the same way. Mirroring that shape would close it.

### 2. `normalizedRoots` already exists — non-blocking

The same doc section asks for root folding via `normalizedRoots`, and the Claude path uses it (`LocalUsageReader.swift:88`). `codexRolloutFiles(in roots:)` introduces a separate `byPath` dictionary instead. The existing helper resolves symlinks and collapses nested roots; raw path-string comparison does neither.

With two hardcoded sibling directories this makes no practical difference today, which is why it's a note rather than a request.

### 3. The source links don't resolve — note

`rollout/src/lib.rs#L600-L603` points past the end of that file (144 lines); the constants are at L67-68. Branch links drift, so a commit-pinned permalink would keep the root-cause evidence readable later.

### Verified as correct

- The relocation test is real: narrowing the scan to `[root]` fails it with `[]` vs `[110]`, so it reads the archived file rather than passing on a stale cache blob.
- `FileManager.enumerator` recurses, so a flat or date-nested archive layout both work.
- `codexRoot` / `codexRoots` mirrors the existing `claudeRoot` / `claudeRoots` pair, and a single `codexScanRoots` list matches the extension convention.

ite476:
@chattymin 

Thanks for the thorough review. I addressed the blocking test gap and incorporated the two follow-up notes.

- Added `computeCodexScanRoots(home:)` so the active and archived Codex roots are defined in one testable place.
- Kept `codexScanRoots` as the shared source used by the reader and cache.
- Added regression coverage for the injected default roots and symlinked roots.
- Updated multi-root Codex scanning to use `normalizedRoots`.
- Updated `docs/reference/provider-extension.md` to document the Codex active/archive root rule.
- Replaced the PR's upstream Codex references with commit-pinned permalinks.

Validation:

- `swift test --filter LocalUsageReaderTests` — 56 tests passed.
- `./scripts/test-gate.sh` — 680 tests passed, 7 skipped, 0 failures.
- Logic-core line coverage: 89.67%.

The changes are pushed in commit b9f369c.

chattymin:

PR #184, opened 2026-08-18
perf: bound memory while parsing large Codex rollouts

poketopa:
## Summary

Bound Codex rollout parsing memory independently of the rollout file size.

- Replace the whole-file `String(contentsOf:)` read with a 1 MiB `FileHandle` line stream.
- Use byte markers to JSON-decode only `session_meta`, model, and `token_count` records.
- Drain `FileHandle`'s bridged `Data` backing storage once per chunk, in addition to the existing per-record JSON autorelease pool.
- Add an opt-in 512 MiB performance regression test and a reproducible local gate.

This is a focused follow-up to #94. That PR bounded autoreleased JSON objects but intentionally retained whole-file reads after a generic streaming experiment regressed memory. This PR addresses a different scale case: a single multi-GiB Codex rollout.

## Root cause

A real 21 GiB Codex rollout made the installed app approach 20 GiB of physical footprint while keeping a CPU core busy. The failure chain was:

1. `parseCodexRollout` decoded the entire JSONL file into one `String`.
2. Splitting that string created additional file-sized storage before line-level parsing began.
3. The per-line autorelease pool from #94 could release `JSONSerialization` objects, but it could not release the whole source string or its split storage.
4. A first chunked implementation still accumulated roughly one file's worth of RSS because `FileHandle.read(upToCount:)` bridged backing storage remained autoreleased until the outer pool drained. A chunk-level autorelease boundary reduced parse growth from 516 MiB to 5 MiB on the 512 MiB fixture.
5. Existing tests used small semantic fixtures, so token correctness was covered but file-sized RSS growth had no regression guard.

## Behavior preservation

The change leaves rollout ownership and accounting after parsing unchanged:

- session and model discovery retain their existing precedence;
- cumulative/last usage vectors, repeated-state normalization, fork replay trimming, and subagent ownership are unchanged;
- file read failures still discard the rollout rather than returning a partial result;
- all existing Codex reader and cache tests pass.

## Performance

Release build, same generated 512 MiB Codex rollout with 64 usage events:

| | Whole-file parser | Streaming parser |
| --- | ---: | ---: |
| Peak RSS | 1,052 MiB | 28 MiB |
| RSS growth during parse | not separately recorded | 5 MiB |
| Parse time | 34.156 s | 2.798 s |

The opt-in test enforces:

- peak RSS below 700 MiB;
- parse-time RSS growth below 192 MiB;
- elapsed time below 20 seconds;
- exact event count and input/cache-read/output/total token sums.

Run it with:

```bash
./scripts/perf-codex-large-rollout.sh
```

The large fixture is generated at runtime and is not committed.

## Class sweep and scope

I searched the remaining usage readers for whole-file reads. Claude, Gemini, and Grok still have provider-specific whole-file paths. They are intentionally unchanged here:

- only Codex reproduced a single multi-GiB rollout on the measured machine;
- #94 documented that a generic streaming conversion regressed settled memory;
- changing every provider would broaden the token-accuracy risk without matching provider-specific evidence.

This PR also does not add incremental append parsing. A changed large rollout still requires a cold sequential scan; the goal here is to bound memory and remove unnecessary JSON parsing. Incremental checkpoints can be evaluated separately.

Open PR #181 changes Codex scan roots but not rollout parsing. This parser change is orthogonal and should compose after a rebase if #181 lands first.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Other: performance

## Checklist

- [x] No UI changes

## Validation

- `./scripts/perf-codex-large-rollout.sh`
  - 512 MiB: baseline RSS 22 MiB, peak RSS 28 MiB, growth 5 MiB, 2.798 s
  - 84 targeted `LocalUsageReaderTests` / `LocalUsageCacheTests`, 0 failures
  - full suite: 677 tests, 8 skipped, 0 failures
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer ./scripts/test-gate.sh`
  - 677 tests, 8 skipped, 0 failures
  - logic-core line coverage: 90.01% (75% required)
- `git diff --check`
- `bash -n scripts/perf-codex-large-rollout.sh`

### DIFF
--- a/Sources/PokeTokenBar/Core/LocalUsageReader.swift
+++ b/Sources/PokeTokenBar/Core/LocalUsageReader.swift
@@ -386,7 +386,7 @@ enum LocalUsageReader {
 
     /// 파일 내부 정보만 파싱. fork replay 여부는 다른 rollout과 대조.
     static func parseCodexRollout(_ url: URL, fmt: DateFormatter) -> CodexParsedRollout {
-        guard let text = try? String(contentsOf: url, encoding: .utf8) else {
+        func emptyRollout() -> CodexParsedRollout {
             return CodexParsedRollout(
                 path: url.path,
                 sessionID: nil,
@@ -396,6 +396,7 @@ enum LocalUsageReader {
                 events: []
             )
         }
+
         var events: [CodexUsageEvent] = []
         var turn = 0
         var sessionID: String?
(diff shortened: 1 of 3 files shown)


chattymin:

Issue #149, opened 2026-08-07
Add support for a new UI language — architecture & contribution guide


chattymin:
## What this is

PokeTokenBar currently ships in three languages — English, 한국어, 日本語. This is a
**standing invitation** to add a new UI language, and the guide for doing it without
bending the app's localization architecture.

Want to add a specific language? Either:
- **Open a new issue that references this one** (e.g. `Add French (fr) localization`), so each language can be tracked separately, **or**
- **Open a PR directly** that follows the checklist below and links back here.

## How localization works (the places you touch)

All UI strings live in code — there are no `.strings` / `.xcstrings` files. Adding a
language is compiler-guided end to end.

1. **Add a case to `AppLanguage`** — `Sources/PokeTokenBar/Core/CompanionModel.swift:9`
   ```swift
   enum AppLanguage: String, Codable, Sendable, CaseIterable {
       case ko, en, ja            // + your language code
       var apiCodes: [String]     // PokéAPI `language.name` candidates for Pokémon names
       var label: String          // native-script name shown in Settings, e.g. "Français"
   }
   ```
   The Settings picker iterates `AppLanguage.allCases`
   (`Sources/PokeTokenBar/UI/SettingsView.swift:107`), so the new language shows up
   with **no view changes**. Also add a system-locale mapping to
   `AppLanguage.systemDefault` (`CompanionModel.swift:31`) so fresh installs on that
   locale start in your language — existing users keep their saved choice.

2. **Extend `L.t(...)` and translate the strings** —
   `Sources/PokeTokenBar/Core/Localization.swift:10`
   ```swift
   private func t(_ ko: String, _ en: String, _ ja: String) -> String {
       switch lang { case .ko: return ko; case .en: return en; case .ja: return ja }
   }
   ```
   Add a parameter for your language. Every UI string in the app (popover, settings,
   notifications, floating pet) goes through `t`, so once the signature changes **the
   compiler points at every call site that still needs a translation**. That's the bulk
   of the work: ~200 strings, no hidden string tables to hunt for.

3. **Translate the 25 Pokémon natures** — `PokemonNature.name(_:)`
   (`Sources/PokeTokenBar/Core/CompanionModel.swift:320`). These have official
   translations in the mainline games — use those, don't invent new ones.

4. **(Only if PokéAPI has your language) extend the species-name fetch** —
   `Sources/PokeTokenBar/Core/PokeAPIClient.swift:24`
   ```swift
[...]

PR #185, opened 2026-08-18
feat: add French (fr) UI language

tchapuis:
Add French as a fifth UI language, following the architecture guide in #149 and the Spanish PR (#159) as the template.

- AppLanguage gains a .fr case (apiCodes ["fr"], native label "Français"), and systemDefault maps the fr system locale to it.
- L.t(...) is extended to five languages; every UI string in Localization.swift is translated to French, using the informal "tu" register to match the app's playful companion tone.
- PokemonNature.name(_:) returns the 25 official French nature names, verified against PokéAPI's fr `names` data rather than invented.
- PokeAPIClient.langCodes now includes "fr" so French Pokémon species names are fetched (otherwise they fall back to English).
- Weekly limit labels use the compact "Hebdo" so French alert copy stays within the floating-pet speech bubble (PerformanceTests bubble-fit guard).
- Add LocalizationInterpolationTests: a parity guard asserting every language keeps the same \(...) interpolations, since the compiler only enforces the argument count, not that a translation kept its placeholders.

Refs #149

### DIFF
--- a/Sources/PokeTokenBar/Core/CompanionModel.swift
+++ b/Sources/PokeTokenBar/Core/CompanionModel.swift
@@ -7,7 +7,7 @@ enum CompanionStateKind: String, Sendable {
 
 /// 앱 언어. 포켓몬 이름은 PokéAPI 다국어 names 에서 가져온다.
 enum AppLanguage: String, Codable, Sendable, CaseIterable {
-    case ko, en, ja, es
+    case ko, en, ja, es, fr
     /// PokéAPI language.name 후보(첫 매칭 사용)
     var apiCodes: [String] {
         switch self {
@@ -15,10 +15,11 @@ enum AppLanguage: String, Codable, Sendable, CaseIterable {
         case .en: return ["en"]
         case .ja: return ["ja-Hrkt", "ja"]
         case .es: return ["es"]
+        case .fr: return ["fr"]
         }
     }
     var label: String {
-        switch self { case .ko: return "한국어"; case .en: return "English"; case .ja: return "日本語"; case .es: return "Español" }
+        switch self { case .ko: return "한국어"; case .en: return "English"; case .ja: return "日本語"; case .es: return "Español"; case .fr: return "Français" }
     }
 
     var displayLocale: Locale { Locale(identifier: rawValue) }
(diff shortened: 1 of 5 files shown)


chattymin:
Thanks @tchapuis — this is careful work, and I checked it rather than taking it on trust:

- All 25 French nature names match PokéAPI's `fr` data exactly (fetched all 25 and compared; zero mismatches).
- Of 208 `t(...)` strings, only 6 have French identical to English — `Collection`, `Maintenance`, `Notifications`, `\(m) min`, `Pokédex`, `Rare` — and all six are genuinely the same word in French. No untranslated leftovers.
- `swift test` on your head: 677 tests, 0 failures.

`LocalizationInterpolationTests` is the part I'm happiest about. It iterates `allCases`, so it retroactively guards the four existing languages too — that outlives the French itself.

Apologies for the churn: `main` has moved a fair bit since you opened this, and the branch now conflicts.

### Rebase — one conflict, plus five strings that need French

`d16c3ea` (#158, representative Pokémon) landed and added UI strings, so `Localization.swift` conflicts in a single hunk. Rebasing resolves that hunk but leaves five four-argument `t(...)` calls needing a French argument:

```swift
t("대표 포켓몬", "Representative Pokémon", "代表ポケモン", "Pokémon representativo")
t("현재 포켓몬 따라가기", "Follow current companion", "現在のポケモンに合わせる", "Seguir al compañero actual")
t("도감에서 선택…", "Choose in Pokédex…", "図鑑で選ぶ…", "Elegir en la Pokédex…")
t("대표로 설정", "Set as representative", "代表ポケモンに設定", "Establecer como representante")
t("대표", "Representative", "代表", "Representante")   // badge — keep it short
```

You can't miss one: after your change `t` takes five arguments, so anything left at four fails to compile. These are popover/settings labels rather than speech-bubble copy, so the bubble-fit guard doesn't constrain their length.

### One more thing, in the same rebase

The commit carries `Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>`. This repo keeps AI attribution out of the history, and a squash merge carries the branch commit message through to `main` — so please drop that trailer while you're amending.

I have `maintainer_can_modify` here and am happy to do the rebase myself if you'd rather not. But the remaining work is really five translations, and you have a better ear for the register you've used throughout than I do. Either way is fine — just say which.

tchapuis:
Thank you for you message @chattymin ! 

I just rebased my work and fixed everything you said. Hope it's good now !

chattymin:

PR #186, opened 2026-08-18
fix: wait for specific PID instead of process name in update loop (#175)

### LINKED ISSUE #175: The update wait loop matches any instance by process name, so it always times out while a duplicate is running
`UpdateChecker.applyUpdate()` calls `NSApp.terminate(nil)` (`UpdateChecker.swift:66`), which ends the instance whose popover was used. The detached script then waits for the app to be gone (`UpdateChecker.swift:123`):

```sh
for i in $(seq 1 40); do pgrep -x PokeTokenBar >/dev/null 2>&1 || break; sleep 0.5; done
```

`pgrep -x` matches on process name, so it matches *any* instance. With one instance the loop breaks in well under a second, which is the case it was written for. With two, it can never break: the surviving instance keeps the name alive, so the wait always burns its full 20s and falls through.

Two things then follow, both of which the wait was there to prevent (`:116`):

1. `brew upgrade --cask` replaces the bundle underneath a still-running app — the bundle-replacement race the loop was added to avoid.
2. `launchctl kickstart -k` (`:130`) starts a new instance. If the survivor is the launchd job's process, `-k` replaces it and the update comes up. If the survivor is the GUI-launched one — which is the case when the user clicks Update from the *agent* instance's popover — the kickstarted process is the newer of the two, and after #163 it yields. The update does not come up until the next launch.

The population this hits is exactly the one #163 targets: users sitting on a duplicate today. It self-heals, since the bundle on disk is already the new version, so the next launch is correct. Not a blocker, and deliberately out of #163's scope — widening that PR to reach into the updater would trade a small guard for a much larger blast radius.

Fix shape: wait on the terminating process specifically rather than the process name — capture `getpid()` before `terminate` and pass it to the script, then `kill -0 <pid>` in the loop. That keeps the bundle-replacement guard meaningful in both states and leaves the 20s ceiling as a real timeout rather than a guaranteed one.

Class, for `docs/reference/defect-log.md`: **`pgrep -x <name>` is an identity check on the executable, not on the process you are waiting for.** Any wait-for-exit that can run while a sibling of the same binary is alive needs a pid, and a loop that can never break is indistinguishable from a slow quit in the logs.

justinjeong5:
## Summary

Fixes #175

`UpdateChecker.applyUpdate()` triggers `NSApp.terminate(nil)` after launching the detached background update script. The detached script previously waited for the app to terminate by running `pgrep -x PokeTokenBar`. Because `pgrep -x` matches on executable process name, running a duplicate instance caused the wait loop to never break, burning the full 20s timeout and falling through.

This PR updates the detached script to wait on the specific terminating process:

- Passes the terminating process's PID (`ProcessInfo.processInfo.processIdentifier`) as positional argument `$3` to the detached script (`sh brew bundlePath pid`).
- Replaces `pgrep -x PokeTokenBar` with `kill -0 "$3" 2>/dev/null || break` in the wait loop.
- Exposes `detachedUpgradeScript` as `nonisolated static let` for deterministic test verification.
- Adds defect class record to `docs/reference/defect-log.md`: *`pgrep -x <name>` is an identity check on the executable, not on the process you are waiting for.*

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation

## UI changes

No visual UI change. The update script now exits the wait loop immediately upon the terminating instance's exit without unnecessary 20s timeout hangs when duplicate instances exist.

## Checklist

## Tests

`UpdateCheckerTests`:

- `testDetachedUpgradeScriptWaitsOnPidNotProcessName` — asserts `pgrep -x` is not used and `kill -0 "$3"` is present in the wait loop.
- `testDetachedUpgradeScriptUsesPositionalParameters` — asserts all positional parameters (`$1` brew, `$2` bundlePath, `$3` pid) are correctly wired.

`swift test --filter UpdateCheckerTests`: 8 tests, 0 failures.

Full test suite: 278 tests, 0 failures.

## How to verify

```bash
  swift test --filter UpdateCheckerTests
```

---

### DIFF
--- a/Sources/PokeTokenBar/Core/UpdateChecker.swift
+++ b/Sources/PokeTokenBar/Core/UpdateChecker.swift
@@ -110,31 +110,36 @@ final class UpdateChecker {
         return process.terminationStatus == 0
     }
 
-    /// 앱이 완전히 종료된 뒤 tap 갱신 + cask 업그레이드 + 재오픈을 수행하는 분리(detached) 스크립트.
+    /// 앱이 완전히 종료된 뒤 tap 갱신 + cask 업그레이드 + 재오픈을 수행하는 분리(detached) 스크립트 본문.
     /// - `brew update` 선행: auto-update 빈도 제한(기본 24h)으로 stale 한 로컬 tap 때문에 `brew upgrade`
     ///   가 no-op(exit 0) 되어 "업데이트 안 됨 + 앱만 종료"가 나던 문제를 막는다.
-    /// - 앱 종료를 기다림(pgrep): 실행 중 번들 교체 레이스 + 재오픈 LaunchServices(-600) 레이스 회피.
+    /// - 앱 종료를 기다림(`kill -0 $3`): 실행 중 번들 교체 레이스 + 재오픈 LaunchServices(-600) 레이스 회피.
+    ///   `pgrep -x` 대신 특정 PID를 감시하여 중복 인스턴스가 실행 중일 때도 20s 타임아웃 없이 즉시 진행 (#175).
     /// - brew 를 백그라운드+워치독(≤300s)으로 감싸 hang 시에도 reopen 이 반드시 실행되게 함
     ///   (앱이 종료된 채 영영 안 돌아오는 것 방지). 종료 직후 재오픈 실패 대비 `open` 재시도.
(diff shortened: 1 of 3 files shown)


chattymin:

PR #182, opened 2026-08-18
feat: support multi-root discovery for Antigravity 2.0 and IDE

justinjeong5:
## Summary

`LocalAntigravityUsageReader` previously looked only under `~/.gemini/antigravity-cli/conversations/*.db`. In **Antigravity 2.0** and **Antigravity IDE**, conversation SQLite databases are saved under `~/.gemini/antigravity/conversations/` or `~/.gemini/antigravity-ide/conversations/`. Furthermore, modern Antigravity 2.0 sessions omit `chat_start_metadata.created_at` from the Protobuf blob, which caused modern conversation tokens to be silently dropped.

This PR introduces comprehensive support for modern Antigravity 2.0 & IDE environments:

- **Multi-root scanning:** `LocalAntigravityUsageReader.defaultRoots` checks all candidate folders (`antigravity`, `antigravity-cli`, and `antigravity-ide`).
- **Protobuf timestamp & Step timeline reconstruction:** Reconstructs per-record timestamps from the store's `steps` table via `response_id` (1:1 direct correlation), falling back to in-order `execution_id` step timings, and database `mtime` as a last resort, while rejecting invalid/overflow sentinels.
- **Provider unification:** Kept under a single provider tab consistent with Claude Code (CLI & Desktop) and Cursor UX.
- **Safe deduplication:** Turns are deduplicated by the protobuf's unique `response_id` via `LocalUsageReader.dedupKeepMax`, so multi-root discovery never double-counts spend.
- **Backward compatibility:** Existing single-root callers (`defaultRoot`, `scan(root:)`, `entries(modifiedSince:root:)`, cache initializers) remain intact.
- **Crash fix:** Safely dispatch `UNUserNotificationCenter.requestAuthorization` with async `Task` to prevent a MainActor isolation crash (`_dispatch_assert_queue_fail`) when opening the popover for the first time on Swift 6 / macOS 15.
- **Docs & Defect Log:** Updated Data sources tables in `README.md`, `README.ko.md`, `README.ja.md`, and added Antigravity step timestamp correlation entry in `docs/reference/defect-log.md`.

### Design Decision & Grouping Rationale

We considered splitting this work into separate PRs, but decided to **group them into a single cohesive PR** for the following reasons:
[...]

PR #187, opened 2026-08-18
feat: add per-provider additional scan folders (#177)

### LINKED ISSUE #177: Extend additional scan folders to every provider, not just Claude
Every provider resolves its log location on its own, and only Claude can be pointed somewhere else from inside the app. #162 adds a settings field for extra Claude scan folders; this issue tracks extending that to the rest.

## Where each provider looks today

| Provider | Location | Override |
| --- | --- | --- |
| Claude | computed multi-root | `CLAUDE_CONFIG_DIR` + settings field (#162) |
| OpenCode | `~/.local/share/opencode` | `OPENCODE_DATA_DIR` |
| Hermes | `~/.hermes` | `HERMES_HOME` |
| Copilot | `~/.copilot` | `COPILOT_HOME` |
| Grok | `~/.grok/sessions` | `GROK_HOME` |
| Codex | `~/.codex/sessions` | none |
| Gemini | `~/.gemini/tmp` | none |
| Antigravity | `~/.gemini/antigravity-cli/conversations` | none |
| Cursor | `~/Library/Application Support/Cursor/User/globalStorage` | none |

Environment overrides are now read through the login shell (#176), so they work from a GUI-launched `.app`. What remains is that four providers have no override at all, and that an environment variable is an awkward thing to ask a menu bar app's user for.

## Why this is worth doing

The failure mode is silent. A relocated log directory does not produce an error — the total is just lower than it should be, and nothing in the UI suggests a folder was missed. #144 removed this for Claude's known locations, #162 removes it for arbitrary Claude locations, and the same hole stays open everywhere else.

No user report for a non-Claude provider yet, which is why this is not urgent. It is also not evidence of absence: the whole point is that an affected user cannot tell.

## Constraint: a folder must belong to a provider

The obvious shape — one flat list of folders that every reader tries — should **not** be used. Readers are not safely interchangeable on the same directory. `LocalUsageReader.jsonlFiles` already carries an `allowJSON` flag that exists specifically so the Gemini reader's `.json` handling does not pick up Claude's `.meta.json`; that hazard generalizes once arbitrary folders are fed to arbitrary parsers. Cross-parser contamination would show up as inflated or misattributed totals, which is harder to notice than the undercount this is meant to fix.

So each entry has to name its provider. Two workable shapes:

- one list per provider (settings keyed by provider id), or
- one list with a provider tag per entry.

The first fits the existing extension convention better — `CONTRIBUTING.md` already says a usage source is added by implementing `UsageProvider` and registering it, so the natural hook is on that protocol, with each reader consulting its own configured roots.

## Storage

#162 stores its value under a provider-scoped key so this work does not need a migration. Whatever shape is chosen here should keep that property.

## Scope notes

- Root discovery differs per provider (recursive `.jsonl` walk vs. locating a SQLite file inside the directory). The *unit* is a directory in every case, so the setting can stay uniform; interpretation stays with each reader.
- `normalizedRoots` folding and hidden-directory skipping interact badly with user-supplied roots — see the discussion in #162 before wiring a second provider through it.
- The settings UI should not grow one text field per provider. A provider picker plus a single list is likely the right presentation, but that is a design question, not a decision made here.

Borisserz:
## Summary

Every provider resolves its log location on its own, and only Claude could be pointed somewhere else from inside the app. A relocated Codex / Gemini / Cursor / … directory does not error — the total is just lower, and nothing in the UI says a folder was missed.

This is the generalized form of #162, landed on `main` so it does not wait on that PR. The feature, the `CLAUDE_CONFIG_DIR` diagnosis, and defect B all originated in @hannin's #162; this PR takes that review's blocking items and extends the extra-folder list to every provider. Storage is one list per provider (`customScanRoots.<id>`), not one flat list every reader tries. Gemini's `allowJSON` walk would otherwise ingest Claude `.meta.json` and mis-attribute tokens.

Custom roots only *add* to curated defaults. An ancestor extra (`~`, `/`) is dropped so `normalizedRoots` cannot fold a hidden default away — `jsonlFiles` uses `.skipsHiddenFiles`, and scanning `~` silently zeros Claude. That is #162-B, baked in here. `.skipsHiddenFiles` is left alone.

#162 can rebase onto these keys or close. There is no generic `customScanRoots` key, so no migration.

Rebased onto current `main`. After #182, Antigravity has three default roots (2.0/Core, CLI, IDE). `resolvedRoots` unions custom extras onto that full list — a CLI-only literal is `defaultRoots[1]` and would silently drop the other two editions. Codex custom roots go through `computeCodexScanRoots`, so archived sessions (#181) stay in the union.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Advanced had no extra-folder field on `main` (#162's Claude-only field is still on its own branch). | Settings → Advanced → **Provider** picker + one **Additional scan folders** list. Persist on submit, focus loss, provider switch, or collapsing Advanced — not per keystroke. Match count is post-fold, off the main thread, debounced. |

## Checklist

## What this takes from the #162 review

Blocking items from the owner review on #162, applied here so they are not re-opened:

- **A.** `L.t(ko, en, ja, es)` — four arguments, current `main`.
- **B.** Custom roots cannot evict a curated default (`~` and `/` included). Tests actually scan a hidden `.claude/projects` jsonl after that.
- **C.** Key is `customScanRoots.claude_code` (and `.codex`, `.gemini`, …), never `customScanRoots`.

Recommendations 1–4: commit on submit / blur, not keystroke; match count off main thread; count after `normalizedRoots`; compute functions default `customRootsValue: nil`, not `UserDefaults.standard`.

French strings are left to #185 on rebase, per merge order.

## Tests

`CustomScanRootsTests` / `AntigravityUsageTests` — each branch alone:

- `testAncestorCustomRootDoesNotEvictCuratedDefault` / `testClaudeAncestorCustomDoesNotZeroJsonlScan` — the #162-B zero-total. Confirmed by scanning a real jsonl under `.claude/projects`.
- `testRootFilesystemCustomDoesNotEvictCuratedDefault` — `/` is an ancestor of everything; `extra + \"/\"` is `\"//\"` so a naive prefix check would walk the whole disk.
- `testEveryRegisteredProviderConsultsItsOwnCustomScanRootsKey` — source sweep for `storedValue(for: \"<id>\")`, plus `curatedRoots(for:)` non-empty for every registered provider (Settings match count).
- `testNoRootScanReadsEveryDefaultRoot` — one turn in each of the three Antigravity default relative paths under a fake home; `resolvedRoots` + `entries` must see all three. Confirmed red against the CLI-only literal (only `r1`).
- `testEntriesWithoutRootsGoesThroughResolvedRoots` — production no-root path and the cache actor both call `resolvedRoots`.
- `testCodexSessionRootsIncludeArchivedSessions` — custom-root union must not drop `#181`'s archived root.
- `testCodexCacheKeepsSessionIndexAcrossMultipleRoots` — prune after the union, not per root.
- `testCodexCacheFindsParentInASiblingRoot` — parent-closure over the union.
- `testSetCustomScanRootsInvalidatesAdditionalProviderCache` — OpenCode/Cursor/Copilot/Kiro 30s cache would hide a newly added folder.
- `testSettingsUsesProviderPickerAndCommitsOnSubmit` — picker + one list; `customScanDraftOwnerID` so a picker click that blurs the field cannot write Claude paths into `customScanRoots.gemini`.

Full suite: 760 pass, 10 skipped, 0 failures.

## How to verify

```bash
  swift test --filter 'CustomScanRootsTests|AntigravityUsageTests.testNoRootScanReadsEveryDefaultRoot'
```

Manual: Settings → Advanced → pick a provider whose logs are outside the built-in path → paste that folder → submit or click away. The next refresh should pick those sessions up. Pointing Gemini at a Claude folder must not inflate Gemini. Typing `~` must not drop today's Claude total to zero.

Closes #177

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -174,6 +174,8 @@ swift test                   # ユニットテスト
 | `status.claude.com`, `status.openai.com` | プロバイダ障害バナー | statuspage の要約；表示専用 — 設定でオフにできます |
 | `api.github.com` | アップデート確認 | 最新リリースのタグ；起動時とポップオーバーを開いた時 |
 
+ログが**上記の既定パスの外**にある場合は、**設定 → 詳細 → 追加スキャンフォルダ**にそのフォルダを追加します。先にプロバイダーを選んでください — フォルダはそのプロバイダーだけが解析するので、Gemini 欄に Claude のログを指定するとトークンの帰属が壊れます。追加フォルダは既定の場所に*足すだけ*で、置き換えません。
+
 ## プライバシー & 権限
 
 - **オンデバイス。** トークン使用量はローカルの Claude Code・Codex・Gemini CLI・Antigravity・OpenCode・Hermes Agent・Cursor・Grok CLI・Copilot CLI・Kiro CLI データから直接読み取ります。使用量のアップロードも、モデルの推論実行も行いません。
--- a/README.ko.md
+++ b/README.ko.md
@@ -174,6 +174,8 @@ swift test                   # 단위 테스트
 | `status.claude.com`, `status.openai.com` | 프로바이더 장애 배너 | statuspage 요약; 표시 전용 — 설정에서 끌 수 있음 |
 | `api.github.com` | 업데이트 확인 | 최신 릴리스 태그; 기동 시와 팝오버를 열 때 |
 
+로그가 **위 기본 경로 밖**에 있으면 **설정 → 고급 → 추가 스캔 폴더**에 그 폴더를 넣습니다. 프로바이더를 먼저 고르세요 — 폴더는 그 프로바이더만 파싱하므로, Gemini 칸에 Claude 로그를 넣으면 토큰이 잘못 귀속됩니다. 추가 폴더는 기본 위치에 *더해질* 뿐 대체하지 않습니다.
+
 ## 프라이버시 & 권한
 
 - **온디바이스.** 토큰 사용량은 로컬 Claude Code·Codex·Gemini CLI·Antigravity·OpenCode·Hermes Agent·Cursor·Grok CLI·Copilot CLI·Kiro CLI 데이터에서 직접 읽습니다. 사용량을 업로드하거나 모델 turn을 실행하지 않습니다.
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ swift test                   # unit tests
 | `status.claude.com`, `status.openai.com` | provider incident banner | statuspage summary; display only — turn it off in Settings |
 | `api.github.com` | update check | latest release tag; on launch and when the popover opens |
 
+If a provider's logs live **outside** those built-in paths, add the folder in **Settings → Advanced → Additional scan folders**. Pick the provider first — each folder is parsed only by that provider, so pointing a Gemini field at Claude logs would mis-attribute tokens. Extra folders are added to the built-in locations; they never replace them.
+
 ## Privacy & permissions
 
 - **On-device.** Token usage is read directly from local Claude Code, Codex, Gemini CLI, Antigravity, OpenCode, Hermes Agent, Cursor, Grok CLI, Copilot CLI, and Kiro CLI data. The app never uploads usage or runs model turns.
--- a/Sources/PokeTokenBar/Core/CustomScanRoots.swift
+++ b/Sources/PokeTokenBar/Core/CustomScanRoots.swift
@@ -1,0 +1,118 @@
+import Darwin
+import Foundation
+
+/// Per-provider extra scan folders from Settings (#177).
+///
+/// One list per provider, stored under `customScanRoots.<id>`. A flat list shared
+/// across readers is unsafe — `jsonlFiles(allowJSON:)` exists specifically so Gemini
+/// does not ingest Claude `.meta.json`. Custom roots only *add* to curated defaults;
+/// an ancestor extra is dropped so `normalizedRoots` cannot fold a hidden default
+/// away (#162-B: scanning `~` with `skipsHiddenFiles` silently zeros usage).
+enum CustomScanRoots {
+
+    static func defaultsKey(for providerID: String) -> String {
+        "customScanRoots.\(providerID)"
+    }
+
+    static func storedValue(for providerID: String, defaults: UserDefaults = .standard) -> String? {
+        guard let raw = defaults.string(forKey: defaultsKey(for: providerID)) else { return nil }
+        let trimmed = raw.trimmingCharacters(in: .whitespacesAndNewlines)
+        return trimmed.isEmpty ? nil : raw
+    }
+
+    /// Comma/newline split, tilde expansion, per-segment `*`/`?`/`[]` via `fnmatch`.
+    /// Entries point at log roots themselves (no `/projects` suffix). Missing paths
+    /// are not errors — a pattern can be registered before the instance exists.
+    static func expand(_ raw: String) -> [URL] {
+        let fm = FileManager.default
+        var out: [URL] = []
+        for part in raw.split(whereSeparator: { $0 == "," || $0.isNewline }) {
+            let pattern = part.trimmingCharacters(in: .whitespaces)
+            guard !pattern.isEmpty else { continue }
+            let absolute = NSString(string: pattern).expandingTildeInPath
+            guard absolute.hasPrefix("/") else { continue }
+
+            var paths = [""]
+            for segment in absolute.split(separator: "/").map(String.init) {
+                if segment.contains("*") || segment.contains("?") || segment.contains("[") {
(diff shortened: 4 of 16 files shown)


chattymin:
Thanks @Borisserz — a lot of surface handled carefully. I checked the parts that matter rather than reading them:

- I disabled `wouldEvict` and reran: 4 tests go red, including `testClaudeAncestorCustomDoesNotZeroJsonlScan` failing `0 != 1` against a real hidden `.claude/projects` jsonl. That reproduces #162-B through an actual scan rather than a list-shape assertion — the form I asked for.
- `test-gate.sh` adds `CustomScanRoots.swift` to `LOGIC_CORE` instead of routing around the gate.
- Wildcards survive the generalization (`fnmatch` per segment), so nothing regresses against #162.
- README en/ko/ja all updated. Full suite here: 701 tests, 0 failures.

Blocking items A/B/C from the #162 review and recommendations 1–4 are all in.

### 1. Rebase — and one part of it is not mechanical

`main` moved: #179, #181, #182 and #184 all touch the four conflicting files. Most is textual — your own #179 changed `kiroEntries` to return a tuple, but the old overload is still there, so that resolves cleanly.

**#182 is the one to watch.** It landed multi-root discovery for Antigravity:

```swift
static var defaultRoots: [URL] {
    [ ".gemini/antigravity/conversations",      // 2.0 / Core
      ".gemini/antigravity-cli/conversations",  // CLI
      ".gemini/antigravity-ide/conversations" ] // IDE
}
static var defaultRoot: URL { defaultRoots[0] }
```

`resolvedRoots` here hardcodes one default:

```swift
defaults: [home.appendingPathComponent(".gemini/antigravity-cli/conversations")]
```

That is now `defaultRoots[1]` — not even `defaultRoot`. Since this PR rewrites `entries(modifiedSince:)` to go through `resolvedRoots`, resolving the conflict textually drops the 2.0/Core and IDE roots: a silent undercount for those users, which is the class this PR exists to remove. Please union against `defaultRoots` rather than a literal path.

### 2. The tests will not tell you if you miss it

I checked what would catch it. Nothing does:

| test | why it stays green |
| --- | --- |
| asserts on `defaultRoots` directly | this PR does not change `defaultRoots` |
| `defaultRoot == defaultRoots[0]` | unchanged |
| `entries(..., root:)` / `entries(..., roots:)` | roots passed explicitly |

Nothing calls `entries(modifiedSince:)` with no root and asserts all three defaults get scanned. Worth adding that one — it is the assertion that makes `resolvedRoots` load-bearing.

### 3. `curatedRoots` is a second registry, and it is undocumented — non-blocking

`CustomScanRoots.curatedRoots(for:)` switches on provider id, but `testEveryRegisteredProviderConsultsItsOwnCustomScanRootsKey` only sweeps for `storedValue(for:)`, and the `provider-extension.md` paragraph this PR adds never mentions `curatedRoots`. A new provider that follows the doc lands in `default: return []`.

Blast radius is small: the only caller is the Settings match count (`SettingsView.swift:513`), and scanning stays correct because each reader passes its own defaults. But the doc now describes one of the two places a new provider has to touch. Either extend the sweep to cover it, or derive the curated roots from the registered providers so there is only one list.

### 4. Credit — non-blocking

This owes a fair amount to #162: the feature itself, the `CLAUDE_CONFIG_DIR` diagnosis behind it, and the B defect, which I only found while reviewing that PR. A line crediting @hannin in the description would be right. I will close #162 as superseded rather than rejected once this lands.

### Merge order

I am taking this ahead of #185 (French UI), so you do not need to produce French for the five new strings — the French contributor will pick them up on their rebase.

Borisserz:
Rebased onto current `main`.

1. `resolvedRoots` unions custom extras onto `defaultRoots(home:)` — all three editions from #182, not the CLI path (`defaultRoots[1]`). `testNoRootScanReadsEveryDefaultRoot` seeds one turn in each relative path under a fake home; it was red against the literal (only `r1`) and is green after the union. The no-root `entries(modifiedSince:)` path and the cache actor both go through `resolvedRoots`.

2. Same class on Codex: `codexSessionRoots` now unions `computeCodexScanRoots`, so archived sessions (#181) are not dropped when custom folders are set.

3. `testEveryRegisteredProviderConsultsItsOwnCustomScanRootsKey` now also requires `curatedRoots(for:)` non-empty for every registered id. `provider-extension.md` names that second registry.

Credit for the feature, the `CLAUDE_CONFIG_DIR` diagnosis, and defect B is @hannin / #162 — added to the description.

No French strings; #185 can pick them up on rebase.

`swift test`: 760 pass, 10 skipped.

chattymin:

PR #188, opened 2026-08-18
feat: add Traditional Chinese (zh-Hant) localization

squall:
## Summary

Adds Traditional Chinese (`zh-Hant`) as the fifth UI language, alongside ko / en / ja / es.

- `AppLanguage` gains `case zhHant = "zh-Hant"`. The raw value doubles as the persisted save format and as the `displayLocale` identifier, so the popover's relative dates follow the app language.
- All 207 `t(...)` call sites and the 25 nature names get their zh-Hant column. The existing ko/en/ja/es columns are untouched — I verified this mechanically by re-parsing both revisions and diffing the first four arguments of every call (0 differences).
- Nature names use the official Traditional Chinese names from the main series (勤奮 / 怕寂寞 / 固執 …).
- Pokémon names come from PokéAPI's `zh-hant` data at runtime, so no name table is bundled.
- The system-language guess matches Traditional variants only (`zh-Hant`, `zh-TW`, `zh-HK`, `zh-MO`). Simplified users (`zh-Hans`, `zh-CN`, `zh-SG`) keep the English fallback rather than being shown a script they did not ask for, and a bare `zh` is not assumed to be Traditional.
- Adds `README.zh-Hant.md`, linked from the language row of the other three.

**Why `zh-Hant` rather than `zh-TW`:** the script subtag is region-neutral, so one entry serves TW/HK/MO; it matches PokéAPI's own `zh-hant` code, keeping one concept end to end; and it stays consistent with the region-neutral `ko`/`en`/`ja`/`es` entries. A future `zh-HK` for Hong Kong vocabulary can be added later without renaming anything. The wording itself leans Taiwan Mandarin.

### One change worth a second look

`PokeAPIClient.langCodes` was a literal array that has to be edited by hand for every new language, and forgetting it is silent — the new language resolves no species name and quietly falls back to English, which reads like a data gap rather than a missing line of code. I changed it to derive from `AppLanguage.apiCodes`, which also meant widening it from `private let` to `static let` so a test can assert the invariant. The resulting set is a superset of the old list, so lookups are unchanged for the existing languages. It is an isolated commit (`0904bc8`) — happy to drop or split it if you would rather keep that line explicit.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## UI changes

No layout changes. The Settings language picker is `.menu` style and driven by `AppLanguage.allCases`, so it gains one entry with no change to the view code. Every other screen keeps its layout with Traditional Chinese text.

| Before | After |
| ------ | ----- |
| Language picker: 한국어 / English / 日本語 / Español | Language picker: 한국어 / English / 日本語 / Español / 繁體中文 |

`PerformanceTests.testLocalizedAlertBubbleFitsDefaultPanel` already iterates `AppLanguage.allCases`, so the new alert strings are verified to fit the floating pet's speech bubble without truncating.

## Checklist

- [x] `swift build` and `swift test` pass locally — 686 tests, and `./scripts/test-gate.sh` reports 89.93% line coverage on the logic core

### On the tests

The argument count of `t(...)` is compiler-enforced, so a missing string cannot ship. `TraditionalChineseTests` covers what the compiler cannot see: the Traditional/Simplified matching table (including script tags winning over region — `zh-Hant-CN` is Traditional, `zh-Hans-TW` is not), every branch of `resolve(preferredLanguages:)` with the locale injected rather than read from the host machine, that the relative-time formatter produces 小時 and not 小时, that the client keeps an API code for every language, that interpolations survived translation, and script leakage (every sampled string carries Han characters and no Hangul or kana — which is what a copy-pasted column looks like). Both guards were checked by injecting the defect they are meant to catch.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -13,7 +13,7 @@
 [![License](https://img.shields.io/badge/license-MIT-3fb950)](LICENSE)
 [![Sponsor](https://img.shields.io/badge/Sponsor-%E2%99%A5-ea4aaa?logo=githubsponsors&logoColor=white)](https://github.com/sponsors/chattymin)
 
-[English](README.md) · [한국어](README.ko.md) · **日本語**
+[English](README.md) · [한국어](README.ko.md) · **日本語** · [繁體中文](README.zh-Hant.md)
 
 </div>
 
@@ -75,7 +75,7 @@ PokeTokenBar は、あなたがすでに使っている AI コーディングト
 <td width="45%" align="center"><img src="assets/settings-ja.png" width="300" alt="設定"></td>
 <td width="55%" valign="middle">
 <h3>設定はお好みで</h3>
-メニューバー表示項目、更新間隔（1–15分／手動）、ログイン時に起動、上限セクションだけを隠す Keychain オフ、警告／危険の閾値つき上限通知、パートナーのイベント通知。<b>韓国語／英語／日本語</b>の UI とポケモン名を完備。
+メニューバー表示項目、更新間隔（1–15分／手動）、ログイン時に起動、上限セクションだけを隠す Keychain オフ、警告／危険の閾値つき上限通知、パートナーのイベント通知。<b>韓国語／英語／日本語／スペイン語／繁体字中国語</b>の UI とポケモン名を完備。
 </td>
 </tr>
 <tr>
--- a/README.ko.md
+++ b/README.ko.md
@@ -13,7 +13,7 @@
 [![License](https://img.shields.io/badge/license-MIT-3fb950)](LICENSE)
 [![Sponsor](https://img.shields.io/badge/Sponsor-%E2%99%A5-ea4aaa?logo=githubsponsors&logoColor=white)](https://github.com/sponsors/chattymin)
 
-[English](README.md) · **한국어** · [日本語](README.ja.md)
+[English](README.md) · **한국어** · [日本語](README.ja.md) · [繁體中文](README.zh-Hant.md)
 
 </div>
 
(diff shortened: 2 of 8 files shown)


chattymin:
Hi @squall — apologies for how long this has sat. It needs a rewrite rather than a rebase, and I'd rather tell you that clearly than leave it drifting.

This branch was written when the app had four languages: it changes `t(ko, en, ja, es)` to add a `zhHant` column. `main` is now at six — French landed in #185 and Brazilian Portuguese in #215 — and a German PR (#234) is approved and waiting on its own rebase. So the 176 strings here are two columns behind, and conflict across six files.

If you're still up for it, what's needed is to redo it against the current signature. Please base it on `main` *after* #234 merges rather than now, otherwise the same thing happens again — I'll comment here when that lands so you have a stable base.

The parts specific to Traditional Chinese looked good on the read-through: matching zh-Hant/TW/HK/MO while deliberately not matching Simplified is the right call, and persisting the raw value as `zh-Hant` rather than a bare `zh` will age better.

Entirely understandable if you'd rather not redo it — say so and I'll close it with thanks rather than leave it open indefinitely.

chattymin:
Following up as promised: #234 (German) is merged, so `main` is a stable base now at **seven** languages — ko, en, ja, es, fr, pt, de.

If you take this on, `t(...)` needs eight columns with zh-Hant added. One thing worth knowing before you start: I rebased #234 myself rather than sending it back, so the same is on offer here — if you'd rather write only the Traditional Chinese strings and leave the mechanical column-widening to me, say so and I'll do that half.

Also carrying over the one substantive point from the automated review on this PR: the comment above `AppLanguage.apiCodes` says an uppercase tag won't match PokéAPI responses, but the implementation returns both `zh-hant` and `zh-Hant`. Worth settling which is true while you're rewriting it.

And the earlier offer stands — if you'd rather not redo it, just say so and I'll close it with thanks.

PR #193, opened 2026-08-19
fix: prevent invisible text in floating pet hover tooltip

ite476:
## Summary

Fixes invisible text in the floating pet hover tooltip when macOS is using dark mode.

## Root cause

The tooltip text used the dynamic `labelColor`, while the bubble background converted `windowBackgroundColor` to `CGColor` before the view's effective appearance was applied. In dark mode, the text and background could therefore resolve from different appearances.

### Dark mode reproduction

Dark mode: tooltip text is invisible

### Light mode reference

Light mode: tooltip text is visible

## Changes

- Resolve tooltip text, background, and border colors from the same effective appearance.
- Apply the appearance to the hover panel and its container view.
- Add regression coverage for both Aqua and Dark Aqua appearances.

## Tests

- `swift test --filter FloatingPetEnergyTests.testHoverCalloutColorsFollowLightAndDarkAppearance`
- `./scripts/test-gate.sh`
- 702 tests passed, 8 skipped
- Logic-core line coverage: 89.99%

## Scope

This change only affects the floating pet hover tooltip's appearance.

### DIFF
--- a/Sources/PokeTokenBar/UI/FloatingPetPanel.swift
+++ b/Sources/PokeTokenBar/UI/FloatingPetPanel.swift
@@ -46,6 +46,34 @@ final class FloatingPetController: NSObject, NSWindowDelegate {
         var wouldTruncate: Bool
     }
 
+    /// AppKit 호버 콜아웃에 사용할 appearance 해석 완료 색상.
+    ///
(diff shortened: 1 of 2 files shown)


chattymin:

PR #194, opened 2026-08-19
fix: preserve sprite aspect ratio in menu bar and popover

Noahbanderson:
## Summary

Gen-V animated sprites are stretched to fill a square frame, so any species whose
GIF canvas is not square renders distorted. Spoink (#325) has a 36x66 canvas and
comes out **1.83x too wide** in the menu bar, the home card, and the evolution line.

Two render paths force the square:

- `SpriteView` — `.resizable().frame(width: size, height: size)` with no aspect mode
- `AppDelegate.menuBarImage` — `sprite.draw(in: NSRect(width: h - 2, height: h - 2))`

Cursor 2026-08-19 09 54 31

PokeTokenBar 2026-08-19 09 54 44

Worth calling out *why this stayed hidden*, since it shapes the fix: every static
asset is already square (species PNGs 96x96, item PNGs 30x30), so stretch-to-fill
was a no-op on the static path — the only path any test measured. The distortion
could only ever appear on the animated path. Other real canvases: Pikachu #25 is
50x46, Gengar #143 is 74x75.

This routes every render site through a single `SpriteFit.size(for:box:)`:

- **`SpriteView`** sizes the inner image to fit and keeps the outer `size x size`
  slot. The outer slot matters — `EvoLineView.rowWidth` and the dex grid assume
  square cells, so shrinking the outer frame would shift those layouts.
- **`menuBarLayout`** keeps the canvas 22pt tall so the baseline cannot jitter
  between frames, and tracks the fitted width. A fixed 22x22 canvas would pad a
  tall sprite with ~5pt of dead space per side, pushing the usage number away from
  it. Square sources keep the exact previous geometry (22x22 canvas, 20x20 at x=1),
  and the bob offset is unchanged.
- **`ItemIconView`** gets the same treatment. Today's item art is square, but it is
  externally sourced, so non-square input is reachable.

A zero-sized source (undecodable image) falls back to the square box rather than
dividing by zero.

Per the defect protocol in `CLAUDE.md`, the class is recorded in
`docs/reference/defect-log.md` under a new "render geometry" section, including the
"static assets are all square, so only the GIF path shows it" reasoning that made
this hard to catch.

image

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

Menu bar status item, home card sprite, evolution line thumbnails, and dex grid
thumbnails. Only non-square (animated Gen-V) sprites change; square sources render
byte-identically to before.

| Before | After |
| ------ | ----- |
| Spoink's 36x66 GIF stretched to fill a square: the body is visibly wide and flat, the pearl smeared horizontally. | Spoink renders at its source ratio — narrow and tall, pearl round. |

## Checklist

### Test notes

`SpriteAspectRatioTests` (10 cases) uses the real PokeAPI canvas dimensions above and
states the trigger explicitly — *a non-square source must not render square* — so the
suite cannot pass by exercising square inputs alone. Guards were verified by injecting
the old square-stretch behavior into `SpriteFit.size`: 4 fail, including the trigger.

Full suite on this branch: **713 tests, 1 failure**. The failure is
`LocalAdditionalUsageTests.testHermesAcceptsMillisecondStartedAt`, which reproduces
identically on a clean `main` checkout and is unrelated to this change — it looks
timezone-dependent (expects `2026-01-02`, gets `2026-01-01`). Untouched here; happy to
open a separate issue if useful.


chattymin:

PR #198, opened 2026-08-20
Price claude-fable-5 from the updated LiteLLM snapshot

shawnla90:
## Summary

`claude-fable-5` was bundled as `$0` because the model was unpriced in the ccusage/LiteLLM snapshot when the table was built. The current LiteLLM snapshot now lists it (`claude-fable-5`: $10 input / $50 output / $12.50 cache-write / $1.00 cache-read per Mtok, matching Anthropic's published rates), so every fable-5 token currently undercounts the day/week/month cost totals silently — on a fable-heavy setup, by hundreds of dollars per month of API-equivalent cost.

- Update the exact-match row to the snapshot rates.
- Add a `fable` family fallback for version drift, mirroring the existing opus/sonnet/haiku fallbacks. Placed after the `grok`/`antigravity/` short-circuits so those still win (covered by assertions below).
- I verified every other Claude row in the table still matches the current LiteLLM snapshot exactly — only this row was stale.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## Checklist

- [x] UI changes are described above (before/after — images optional) — no UI changes
- [x] No copyrighted assets, secrets, or private tooling references are committed (see CONTRIBUTING)

Test notes: updated the fable-5 expectation in `testPricingExactAndFallbackAndZero` (1M of each token type → $73.50) and added a family-fallback assertion (`claude-fable-6` → $10 input). Local machine has Command Line Tools only (no XCTest), so the suite ran on a fork CI run of this branch; `swift build` verified locally.

### DIFF
--- a/Sources/PokeTokenBar/Core/ModelPricing.swift
+++ b/Sources/PokeTokenBar/Core/ModelPricing.swift
@@ -24,7 +24,7 @@ enum ModelPricing {
         "claude-opus-4-7":            .perMillion(5, 25, 6.25, 0.5),
         "claude-sonnet-4-6":          .perMillion(3, 15, 3.75, 0.3),
         "claude-haiku-4-5-20251001":  .perMillion(1, 5, 1.25, 0.1),
-        "claude-fable-5":             .zero,                          // ccusage 미가격 → $0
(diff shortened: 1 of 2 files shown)


chattymin:

PR #199, opened 2026-08-20
Label official limits with the account they belong to

shawnla90:
## Summary

Two Claude Code accounts on one Mac (e.g. a personal Max plan and an employer Team plan, switched via `CLAUDE_CONFIG_DIR`) overwrite the single `Claude Code-credentials` Keychain item whenever either CLI refreshes its token. The official-limits section then silently flips between accounts: the five-hour/weekly bars belong to whichever account refreshed last, and nothing in the UI says which one that is. (Observed on my own machine — the limits alternated between my Max 5x and my Team plan depending on which CLI ran last, which is what sent me digging.)

This PR labels the limits with the account they actually belong to:

- Fetch the token owner's identity from `https://api.anthropic.com/api/oauth/profile` — same host, auth header, and beta header as the existing `oauth/usage` call. Best-effort: any failure and the limits render exactly as before, with no label.
- Cache the identity per access token (`OAuthProfileCache` actor), so limit polling does not double its request count — one profile request per new token. Failures are not cached, so a stale label from a previous account can never stick after a token switch.
- Parsing is a pure function (`OAuthProfileData.identity(from:)`) tested with fixtures. Email is required; a response without one is discarded entirely rather than producing a partial label.
- `LimitStatus.accountDisplay` renders `email · org name`, suppressing the auto-generated personal org ("<email>'s Organization") since it duplicates the email.
- No new Keychain access and no new secret handling — the profile call reuses the access token already held in memory for the usage call.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Limits section shows `Plan Max 5x` above the five-hour/weekly bars, with no indication which account they belong to | One extra tertiary caption under the plan row: `Account dev@example.com · Acme Corp` (or just `Account dev@example.com` on personal plans). When the profile lookup fails or hasn't run, the section is pixel-identical to before |

New string `limitsAccount` localized for all four languages (ko/en/ja/es).

## Checklist

- [x] No copyrighted assets, secrets, or private tooling references are committed (see CONTRIBUTING)

Test notes: added `testOAuthProfileIdentityParsing` (full/`organization:null`/missing-email/empty-email fixtures) and `testLimitStatusAccountDisplay` (org shown, personal org suppressed, email-only, nil). Local machine has Command Line Tools only (no XCTest), so the suite ran on a fork CI run of this branch; `swift build` verified locally.

Related: this is deliberately labeling-only. Reading multiple credentials / per-account limit selection is a bigger design I'm proposing separately as an issue (multi-account attribution), and per-provider scan folders are already in flight in #162/#187 — this PR doesn't touch scanning at all.

### DIFF
--- a/Sources/PokeTokenBar/Core/Localization.swift
+++ b/Sources/PokeTokenBar/Core/Localization.swift
@@ -53,6 +53,7 @@ struct L {
         }
     }
     func plan(_ p: String) -> String { t("플랜 \(p)", "Plan \(p)", "プラン \(p)", "Plan \(p)") }
+    func limitsAccount(_ a: String) -> String { t("계정 \(a)", "Account \(a)", "アカウント \(a)", "Cuenta \(a)") }
     func forecastReach(_ time: String) -> String {
         t("현재 속도면 \(time) 한도 도달", "At current rate, limit hit at \(time)", "現在のペースで \(time) に上限到達", "Al ritmo actual, límite alcanzado a las \(time)")
     }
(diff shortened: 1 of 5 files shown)


chattymin:

PR #203, opened 2026-08-20
fix: prevent installation requirements text overlap

ite476:
## Summary

- Prevent long provider labels from overlapping their descriptions in the installation requirements panel.
- Wrap long labels within a dedicated column and stack the row on narrow screens.

## Validation

- Reproduced the Korean installation page with the long provider list in a local browser and verified that the columns no longer overlap.
- Verified the installation panel narrow-screen layout at 375px.
- `git diff --check`


chattymin:

PR #208, opened 2026-08-20
Show entry screen instead of flashing setup view at iOS launch

cte13:
## Problem

At cold launch the **Connect to Mac** (IP entry) screen rendered for ~1 second and was then swapped out for the dashboard once the iCloud payload arrived. For anyone with iCloud sync, setup content the user never needed to act on flashed on every launch.

## Root cause

`DashboardView` picked the first screen purely from synchronous state (`host.isEmpty && payload == nil`). That condition is true at launch before the async data-source determination finishes, so `SetupView` rendered first, its own `.task { await store.fetch() }` fired, and the fetch completion invalidated the very condition that rendered the setup screen — replacing it wholesale.

## Fix

- `PhonePayloadStore` gains `hasCompletedInitialFetch`, set in `fetch()`'s `defer` so every exit path (CloudKit success, HTTP success, and all failures) releases the gate.
- `DashboardView` shows a new `LaunchView` entry screen (app name + "Looking for data source…") until the first fetch completes. The boot gate is the first branch, ahead of the setup check.
- Removed the `.task` fetch trigger from `SetupView` — the boot screen now owns the initial fetch, and the existing auto-refresh timer still picks up late-arriving iCloud data while setup is on screen.

Result by path:
- iCloud payload available → entry screen → dashboard (no setup flash)
- No iCloud, no host configured → entry screen → setup (stays)
- Host configured (local HTTP) → entry screen → dashboard or error/retry, as before

## Verification

- `xcodebuild -scheme PokeTokenBariOS -destination 'generic/platform=iOS Simulator' build` — succeeded
- `swift test` in `PokeTokenBarShared` — 6/6 passed
- No regression test: the iOS app target has no test target, so the boot branch can't be exercised under xctest yet. Captured the defect class in `docs/reference/defect-log.md` instead.


cte13:
Closing: created against the wrong repo. Reopening against cte13/PokeTokenBar-iOS.

PR #211, opened 2026-08-21
fix: show the Raising badge only on the current evolution stage

dev-minsoo:
## Summary

This PR updates the Pokédex `Raising` badge so that it appears only on the current evolution stage of the active companion.

Previously, every reached stage without a graduated record was marked as `Raising`. After an evolution, both the previous and current forms could therefore appear to be raised at the same time. The Pokédex still includes all reached species as before; this change only aligns the badge with the companion's current form.

The previous behavior was intentional: it distinguished Pokédex entries contributed only by the active companion, which can disappear before graduation. This PR preserves that underlying persistence behavior. It only narrows the `Raising` badge to the companion's current form, since showing it on previous forms can suggest that multiple Pokémon are being raised at once.

This treats `Raising` as a current-state indicator rather than an indication of whether a Pokédex entry is permanent. Regression coverage includes an evolved companion, a previously graduated line being raised again, and graduated-only entries.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Every reached stage without a graduated record displayed the `Raising` badge, so both the previous and current forms could appear active after evolution. | Only the active companion's current evolution stage displays the `Raising` badge. Previous forms remain in the Pokédex without a status badge. |

**Before**
raising-badge-before

**After**
raising-badge-after

## Checklist


chattymin:

PR #212, opened 2026-08-21
feat: let users pick animation quality for the menu bar sprite and floating pet

LuceteYang:
## Summary

The fps cap raised each frame's delay (`max(floor, delay)`) instead of dropping frames, so the whole loop stretched. A 55-frame Gen-V sprite (2.75s) played over 22s at the 0.4s floor — **1/8 speed** on the menu bar sprite and the floating pet.

**1. `fix:` cap by decimating, not stretching.** `GIFDecoder.capFrameRate` now emits one frame per floor-length span — loop length matches the original, only the frame rate drops (55 → 6 frames at 0.4s). `Timer.tolerance` 0.5× → 0.1×; tolerance only fires late, so 0.5× allowed a further 1.5× stretch. The menu bar's inline `max(0.4, delay)` becomes a named constant, so the cap can't vanish untested.

**2. `feat:` user-selectable frame rate.** Shared by both always-visible surfaces, in all four UI languages:

| Preset | Floor | ~fps | Idle CPU |
| --- | --- | --- | --- |
| Power saver (default) | 0.4s | 2.5 | current behaviour |
| Balanced | 0.2s | 5 | 1.8% |
| Smooth | 0.1s | 10 | 5.1% |

No preset disables the cap, so the idle-wakeup guarantee holds.

> **Decision needed:** default is **Power saver** — today's frame rate, so no existing user's battery profile changes. Shipping Balanced instead is a one-line change in `UsageStore`.

Regression coverage includes playback-speed preservation on a real 55-frame sprite, verified by re-injecting the old `max(floor, delay)` (reports 22.0s / 11.0s loops and fails). `docs/reference/defect-log.md` records the class.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Long sprites played at 1/8 speed on the menu bar and the floating pet. Frame rate not adjustable. | Original speed at any cap. Settings gains an **Animation** row (Power saver / Balanced / Smooth). Popover unchanged. |

**Before / After** — Zubat (#41), 55 frames × 0.05s, both at the same 0.4s cap:

fix-before-after

**Presets** — the same sprite at each preset, all playing at the correct speed:

presets

<sub>The clips are rendered from the actual sprite and the two cap implementations at 3× scale; the menu bar draws the same frame timing at 22px.</sub>

## Checklist

- [x] `swift build` and `swift test` pass locally (`./scripts/test-gate.sh`: logic-core coverage 89.93% ≥ 75%)
- [x] No copyrighted assets, secrets, or private tooling references are committed (see [[CONTRIBUTING](https://claude.ai/CONTRIBUTING.md)](../CONTRIBUTING.md))


chattymin:

PR #214, opened 2026-08-21
Add oh-my-pi (omp) usage provider

comigor:
## Summary

Adds oh-my-pi (`omp`, a fork of the pi coding agent) as a 12th local usage source, alongside the Pi provider from #189. omp stores the same pi-format session JSONL under `~/.omp/agent/sessions/<cwd>/<timestamp>_<uuid>.jsonl`; every assistant `message` event carries a full `usage` object (`input`/`output`/`cacheRead`/`cacheWrite`/`totalTokens`/`cost`).

Aggregation rules (validated against pi-session-manager's reference implementation and real session files):

- Mirrors `parsePiFile`: every assistant `message` `usage` event plus `compaction`/`branch_summary` envelope usage is summed, aborted/error turns are skipped, total-only usage keeps its aggregate, and one line is one billed API call, so rewound branches count and no dedup is needed beyond per-message ids.
- Subagent session files (`<id>/__advisor.jsonl`, `<id>/<Name>.jsonl`) are included: their tokens are **not** folded into the parent session's usage (unlike Grok's accounting), so excluding them would undercount.
- `bridge/` directories are **excluded**: they hold session *conversion copies* (e.g. written by pi-session-manager) whose usage is already counted at the original source (claude_code / another omp root). Verified against a real bridge file mirroring a `-Projects/` session with the same id.
- Differences from the Pi provider: omp keeps the source-reported `usage.cost.total` via `explicitCost` when positive (real charge pi computed from model pricing; free-tier models report 0 and fall through to the price table, which yields `.zero` for unknown `org/Model` names) and scopes dedup ids by file name, because omp message ids are 8-hex and only session-unique.
- Roots: `~/.omp/agent/sessions` plus `$OMP_CODING_AGENT_DIR/sessions` when set (verified against the omp binary's env var table — unlike pi there is no `*_SESSION_DIR` var). Registered in `CustomScanRoots` and `UsageEnvironmentTests`, so extra scan folders work for omp too.

Registration follows the extension contract: one `UsageProvider` implementation + one entry in `UsageStore.init`'s `providers:` array; scanner, cache, Settings scan folders share the single root definition; no provider-id branches in generic paths. Own `ompParserVersion` cache-invalidation constant, mirroring `piParserVersion`.

Also included: `testHermesAcceptsMillisecondStartedAt` hardcoded a local day that only holds at UTC+0 or east of it (fails in UTC-3); the assertion is now computed from the same instant via `localDayFormatter`, preserving the ms→s conversion check.

## Type of change

- [x] New feature

## UI changes

No `Sources/PokeTokenBar/UI/` changes. Users with omp sessions get an extra `OMP` capsule in the existing provider tab bar (rendered generically from snapshots) and an oh-my-pi entry in Settings → Additional scan folders; omp tokens/cost join today's totals.

## Checklist

- [x] `swift build` and `swift test` pass locally (796 tests, 0 failures; `./scripts/test-gate.sh` logic-core coverage 90.69% ≥ 75%; new branches verified with `llvm-cov --show-regions`, no `^0` regions)
- [x] Tests were added or updated for this change (`OmpUsageTests` — 14 tests incl. real-logs smoke; parity + tab-layout + env-registry registries updated; bridge-exclusion guard defect-injection verified)


chattymin:

PR #215, opened 2026-08-21
feat: add Brazilian Portuguese (pt) UI language

williansaez:
## Summary

Adds Brazilian Portuguese (`pt`) as a fifth UI language, following #159 (Spanish) and #185 (French) as templates.

- `AppLanguage` gains a `.pt` case (native label "Português"), and `systemDefault` maps the `pt` system locale to it.
- `L.t(...)` is extended to five languages and all 212 UI strings in `Localization.swift` are translated, in the informal "você" register that matches the app's playful companion tone.
- `PokemonNature.name(_:)` gains 25 Portuguese names.
- Adds `LocalizationInterpolationTests`, a parity guard asserting every language keeps its `\(...)` placeholders.

Every `ko`/`en`/`ja`/`es` string is byte-for-byte unchanged: the new argument was appended mechanically and the result verified by stripping the fifth argument back off and diffing against `main`.

### Translation decisions worth a reviewer's attention

**Pokémon species names deliberately stay English.** PokéAPI has no `pt` language, and the core series has never shipped a Portuguese localisation, so Brazilian players use the English names anyway — `AppLanguage.resolveName`'s English fallback is the expected result here, not a gap. `"pt"` is still listed in `apiCodes`/`langCodes` so that if PokéAPI ever adds it, it is picked up with no branch changes.

**Nature names are not official, unlike the other four languages.** With no Portuguese core-series localisation there is no official set to copy. They are adjectives agreeing with the feminine "natureza" (as the Spanish ones agree with "naturaleza"), picked so all 25 stay distinct.

**"Shiny" is kept untranslated, on purpose.** Pokémon GO's pt-BR does use an official "Brilhante", but "shiny" is overwhelmingly what the Brazilian community says and this app's audience is Brazilian developers. Flagging it explicitly because it diverges from the Spanish decision to adopt the official "Variocolor" — happy to switch the whole cluster (`Amuleto Brilhante`, `✨ Pokémon Brilhante!`, …) if you prefer consistency with `es`.

Elsewhere the translation does follow official pt-BR franchise terminology where it exists: "Bolsa" for Bag, "chocar" for hatch, "Incubadora"/"Incubando", "Doce Raro" for Rare Candy, "Estágio" for evolution stage, "companheiro" for companion, and the TCG's "Comum/Incomum/Raro".

**Alert copy is length-bounded.** `notifBody` and every window label it embeds are kept no longer than the Spanish equivalent, so the floating-pet speech bubble guard in `PerformanceTests` keeps holding.

**Overlap with #185.** That PR proposes an equivalent interpolation-parity guard under the same filename. Whichever lands first, the other copy can simply be dropped — the guard is not specific to either language.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

No files under `Sources/PokeTokenBar/UI/` are touched. The user-visible change is entirely driven by `Localization.swift`:

| Before | After |
| ------ | ----- |
| Settings → Language offers 한국어 / English / 日本語 / Español. A system locale of `pt` falls back to English. | A fifth entry, **Português**, is offered, and a `pt` system locale selects it on a fresh install. Selecting it renders every screen (popover, Pokédex, catch log, Bag, Shop, Settings, floating-pet bubbles, system notifications) in Brazilian Portuguese. Pokémon species names stay English, as described above. |

## Checklist

- [ ] `swift build` and `swift test` pass locally

> On the first box, honestly: this was authored on Linux, where the macOS-only target cannot be built, so `swift build` / `swift test` were **not** run locally. What was verified instead, mechanically: the `ko`/`en`/`ja`/`es` round-trip diff described above, `\(...)` interpolation parity across all five languages, that all 25 nature names are distinct and non-empty, and that every `L` member called from the new test matches its real signature. CI is the first real compile — please treat a red run as mine to fix.


chattymin:

Issue #200, opened 2026-08-20
Plan-aware cost: label API-equivalent cost and show subscription leverage


shawnla90:
## Problem

The cost figures (today/week/month) are API-equivalent: `ModelPricing` × tokens. That's the right baseline, but for subscription users it answers a different question than the one they ask the menu bar: "what is this actually costing me?"

On a Max or Team plan the marginal cost of a token is $0 — the user pays a flat monthly price. Today the app shows a subscription user a dollar figure that looks like a bill but isn't one. For people using PokeTokenBar to *learn* token economics (a real use case — it's a great teaching tool), the missing piece is the relationship between the two numbers: "you paid $100 this month and burned $610 of API-equivalent tokens — 6.1× leverage."

The app already knows the plan: `subscriptionType` / `rateLimitTier` ride on `LimitStatus` from the cached credential, and the `reportsCost` flag already encodes the "flat-rate providers shouldn't show fake dollars" philosophy for Codex/Cursor/Copilot. This proposal extends that philosophy to Claude subscriptions instead of leaving them priced as if they were API accounts.

## Proposal

1. When the Claude credential reports a subscription plan (`max`/`pro`/`team`), label the existing cost figures as API-equivalent rather than presenting them bare (e.g. `$610.30 (API-equiv.)` or a caption under the cost rows).
2. Optional user-entered monthly price in Settings (default off, no hardcoded price list — plan prices change and vary by seat count/region). When set, show effective leverage for the calendar month: `plan $X/mo · API-equiv $Y · Z×`.
3. API-key accounts (no subscription in the credential) keep exactly today's presentation.

Deliberately **not** proposing: hardcoded plan price tables (stale-prone, per-seat variance), or per-plan zeroing of the cost columns (the API-equivalent number is the teaching value — hiding it would be a regression).

## Design questions

- Is a Settings field for "my monthly plan price" acceptable, or would you prefer label-only (no user-entered price, no leverage ratio)?
- Should the label live on the popover cost rows only, or also affect the menu-bar cost display mode?

Happy to implement once the shape is agreed.

Borisserz:
Taking the **label-only** slice from the proposal (item 1) so Max/Pro/Team popover `$` figures read as API-equivalent, without a Settings price field or a menu-bar change — those are the two open design questions.

PR: https://github.com/chattymin/PokeTokenBar/pull/224

[...]

PR #224, opened 2026-08-23
feat: label popover cost as API-equivalent on Max/Pro/Team (#200)

Borisserz:
## Summary

- When the Claude credential reports a **Max / Pro / Team** plan, popover cost figures (`$610.30`) now read `$610.30 (API-equiv.)` so they are not mistaken for a bill. The number is unchanged — it is still `ModelPricing` × tokens.
- API-key accounts, Free, and providers that already set `reportsCost = false` (Codex / Cursor / Copilot / …) keep today's presentation.
- **Out of scope** (the two design questions on #200): Settings field for monthly plan price / leverage `Z×`, and the menu-bar compact cost. Happy to follow up once those are decided.

Predicate lives on `LimitStatus` (where `subscriptionType` already lives) so generic today/week/month totals do not grow a `providerID ==` branch.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Popover today / week / month / Claude row: `$610.30` on Max/Pro/Team | Same figures with `(API-equiv.)` (localized). API-key and Free unchanged. Menu bar `$9.5` unchanged. |

## Checklist

## Tests

- `LimitStatus.labelsCostAsAPIEquivalent` — max/pro/team (incl. case fold) on; free/api/empty/nil off. A=false and B=true both locked.
- `TokenFormatter.cost(_:labeled:)` — suffix appended; nil/empty suffix does not add a trailing space.
- `UsageStore.labelsCostAsAPIEquivalent` — follows the injected Claude plan after refresh (Max/Team on, Free and no-limits off).
- Source lock: popover consults the predicate + `L.apiEquivalentCaption`; menu-bar `costCompact` stays unlabeled.
- Caption contains `API` and is parenthetical in every `AppLanguage.allCases` (ko/en/ja/es/fr/pt).

`./scripts/test-gate.sh`: **786 pass, 10 skipped, 0 failures**. Logic-core line coverage **90.52%**.

## How to verify

On a Max/Pro/Team Claude login with official limits loaded, open the popover: today / week / month (and the Claude row) should show `(API-equiv.)` next to `$`. Switch the language in Settings — the qualifier localizes; the `$` figure does not. Menu bar cost toggle should still be the compact `$9.5` / `$311` form.

Label-only slice of #200 (Settings price / leverage and menu-bar labeling left for a follow-up).

### DIFF
--- a/Sources/PokeTokenBar/Core/Localization.swift
+++ b/Sources/PokeTokenBar/Core/Localization.swift
@@ -123,6 +123,8 @@ struct L {
     var menuBarItems: String { t("메뉴바 표시 항목 (복수 선택)", "Menu bar items (multi-select)", "メニューバー表示項目（複数選択）", "Elementos de la barra de menús (selección múltiple)", "Éléments de la barre des menus (sélection multiple)", "Itens da barra de menus (seleção múltipla)") }
     var todayTokensShort: String { t("오늘 토큰", "Today's tokens", "本日のトークン", "Tokens de hoy", "Tokens du jour", "Tokens de hoje") }
     var todayCost: String { t("오늘 비용 ($)", "Today's cost ($)", "本日のコスト ($)", "Coste de hoy ($)", "Coût du jour ($)", "Custo de hoje ($)") }
+    /// Max/Pro/Team 팝오버 비용 접미사 — 숫자가 API 환산가임을 표시. 짧아야 캡션 폭에 들어간다.
+    var apiEquivalentCaption: String { t("(API 환산)", "(API-equiv.)", "(API換算)", "(equiv. API)", "(éq. API)", "(equiv. API)") }
     var limitPercent: String { t("한도 %", "Limit %", "上限 %", "Límite %", "Limite %", "Limite %") }
     var limitDisplayModeLabel: String { t("한도 표시 방식", "Limit display", "上限の表示", "Visualización del límite", "Affichage de la limite", "Exibição do limite") }
     var limitDisplayUsed: String { t("사용량", "Used", "使用量", "Usado", "Utilisé", "Usado") }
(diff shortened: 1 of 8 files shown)


chattymin:
Thanks for implementing this, and for taking the conservative half of #200 — label only, no user-entered plan price, no hardcoded price table. That was the right slice to try first.

I'm not taking it, for two reasons.

**The label lands on providers it isn't true for.** The predicate reads the Claude subscription, but `costText` is applied to every provider row and to the combined header. Six providers report cost — Claude, Codex, Gemini, Grok, OpenCode and Hermes — so a user on Claude Max who also uses Grok would see "(API-equiv.)" on the Grok row, where the figure is a server-reported actual charge rather than an estimate. The qualifier would be saying the opposite of the truth there. If this ever comes back, the judgement has to be per-provider — is *this* provider's cost an estimate or a bill — rather than one global flag.

**The part that motivated #200 isn't in it.** The issue's argument was the leverage ratio — "$100/mo against $610 of API-equivalent tokens, 6.1×" — which is what makes the number instructive for someone learning token economics. Without it we pay the UI cost (a caption on every cost row, seven localization columns to maintain) and don't get the payoff. Meanwhile the bare number is the convention in this category and subscription users already read it as "what this would have cost on API".

I'm closing this and leaving #200 open — the underlying observation is fair, it just needs a different shape.

Thanks again, and sorry for the slow answer on this one.

Borisserz:
Follow-up for the two close reasons is in a new PR, not a reopen of this branch:

https://github.com/chattymin/PokeTokenBar/pull/249

Per-provider `costIsEstimate` (Grok/OpenCode/Hermes stay bills). Qualifier only on the leverage row. Optional Settings plan price. `$Y` is estimates only, so Max+Grok does not label Grok as API-equivalent.

Issue #201, opened 2026-08-20
Multi-account attribution: tag usage with the account each Claude config root belongs to


shawnla90:
## Problem

Two Claude Code accounts on one Mac are common (personal Max plan + employer Team plan, switched via `CLAUDE_CONFIG_DIR`). Today this breaks in two independent, silent ways:

1. **Counting:** only the default roots are scanned, so the second account's config dir (e.g. `~/.claude-tradeengage`) is invisible — its tokens never reach the egg. #177 / #187 / #162 address *finding* the folders; this issue is about what happens after they're found.
2. **Attribution:** even once a second root is scanned, everything merges into one undifferentiated total. The transcripts themselves carry no account/org field on any line, so a jsonl parser cannot attribute usage — the *root* is the only attribution boundary that exists on disk.

Real-world numbers from my machine: default root 1,334 session files / 701MB (personal Max 5x), second root invisible — including sessions from the same evening missing from `todayTotalTokens`.

## Proposal

Layered on top of per-provider scan folders (#177/#187), not competing with it:

1. **Root → account mapping:** each Claude config root contains `.claude.json` with an `oauthAccount` block (email, org name, org type). Read it once per root (cached by mtime) and tag the root with that identity.
2. **Entry tagging:** add an optional `accountKey` (the root it came from, resolved to the account identity when known) to `LocalUsageReader.Entry` / the provider snapshot. Dedup stays global by entry ID.
3. **Display:** default UI unchanged (one merged total — the companion aggregates everything, per the no-provider-branch rule). Add a per-account breakdown in the popover's Claude section when more than one account is present, mirroring how `scopedLimitEntries` extends the limits list.

This composes with #187's per-provider folders: those PRs contribute the roots; this contributes what the roots *mean*. It also complements #199 (account labeling on the official limits), which covers the OAuth side of the same two-account confusion.

## Open questions

- Is an account dimension on `Entry` acceptable, or would you rather keep `Entry` flat and aggregate per-root at the provider-snapshot level? (Snapshot-level is smaller; entry-level survives cache merging.)
- `.claude.json` is large (can be several MB) — proposal reads only the `oauthAccount` object via `JSONSerialization` on a mtime-gated cache. Any concerns?

Happy to implement after alignment — flagging the design first since this touches `Entry` and the cache format.

shawnla90:
Design caveat I hit on my own machine after filing this — worth baking into the proposal:

[...]

PR #228, opened 2026-08-26
fix: pick up in-place Claude account switch on auto-poll (#227)

### LINKED ISSUE #227: [BUG] claude code with multiple team account
Hi, usually during my work I switch between two claude code accounts (two mails in the same Team plan) due to reaching the 5h limit. The poketokenbar doesn't recognise the account switch so it stops to display correct usage (while the exp systems seems to level the pokemon).

Borisserz:
## Summary

- Auto-poll now re-reads `~/.claude/.credentials.json` **before** serving the in-memory OAuth cache. `/login` to a second Team email rewrites that file with a new still-valid token; the old cache-until-`expiresAt` path kept serving the previous account's 5h/weekly bars (and the #199 account label) while companion EXP kept moving from local jsonl. Fixes #227.
- Same class in the Antigravity token cache: file tokens are stored with `expiresAt = nil`, so cache-until-expiry never refreshed after `~/.gemini/jetski-standalone-oauth-token` changed.
- Auto-poll stays Keychain-free. A missing file or `"claudeAiOauth": null` keeps the cached token (logout / leftover mcpOAuth-only file under the default path — the case `credentialsFileIsAccountOAuthMissing` already carves out for `CLAUDE_CONFIG_DIR`).

Out of scope: tagging local jsonl by account (#201), Keychain-only switches with no credentials file (still the Refresh button).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## Checklist

- [x] No UI changes

## Tests

`CredentialSwitchCacheTests` calls production `accessToken` with an injected temp file (not a helper copy):

- In-place switch: file A then file B, both unexpired → B token and `planInfo` follow the new file. Restoring the cache-before-file early return fails this test (injected).
- File disappears after a successful load → cached token, Keychain query count 0.
- `"claudeAiOauth": null` after a successful load → cached token (leftover-file case).
- Antigravity twins: file switch + file-gone keep-cache.

`KeychainAutoPathTests` still 0 queries on the automatic path.

`./scripts/test-gate.sh`: **787 pass, 10 skipped, 0 failures**. Logic-core line coverage **90.46%**.

## How to verify

1. Sign into Claude Code account A so `~/.claude/.credentials.json` has a valid token. Open the popover — official 5h/weekly bars (and the account label) match A.
2. `/logout` then `/login` as account B (same Team, different email). Wait one auto-poll (or the existing interval).
3. Official bars and the account label should now match B. Companion EXP should keep moving as before.

Keychain-only installs (no credentials file) are unchanged: press Refresh after a switch.


chattymin:

PR #231, opened 2026-08-27
fix(claude): scope Claude Code keychain lookup by macOS username (#229)

### LINKED ISSUE #229: Ambiguous keychain lookup for `Claude Code-credentials` causes persistent "re-login required" error
### Summary

In `OAuthLimitsProvider.swift` (`readClaudeKeychain`), the keychain query for Claude Code credentials searches solely by `kSecAttrService: "Claude Code-credentials"` with `kSecMatchLimitOne`, without specifying `kSecAttrAccount`. 

When multiple generic-password items share the service name `Claude Code-credentials` in macOS Keychain (for example, a current username-scoped item and a stale or MCP-scoped item like `unknown`), `SecItemCopyMatching` returns an arbitrary item. If the returned item does not contain `claudeAiOauth`, PokeTokenBar reports `LimitsError.credentialMissingAccountOAuth` ("re-login required"), which cannot be resolved by logging in again.

---

### Location

- `Sources/PokeTokenBar/Core/OAuthLimitsProvider.swift` (`readClaudeKeychain`)

```swift
var query: [String: Any] = [
    kSecClass as String: kSecClassGenericPassword,
    kSecAttrService as String: OAuthCredentialData.claudeKeychainService,
    kSecReturnData as String: true,
    kSecMatchLimit as String: kSecMatchLimitOne,
]

### PR DESCRIPTION
**Pull request by rel94**
## Summary

Fixes #229.

Scopes the Keychain generic-password query for Claude Code credentials (`Claude Code-credentials`) by `kSecAttrAccount: NSUserName()`.

### Root Cause
Previously, `readClaudeKeychain` in `OAuthLimitsProvider.swift` queried solely by `kSecAttrService: "Claude Code-credentials"` with `kSecMatchLimitOne`, without specifying `kSecAttrAccount`. If multiple items existed under that service in Keychain (e.g. current macOS user item vs a stale or MCP-only item with account `unknown`), `SecItemCopyMatching` could arbitrarily pick the non-account item, throwing `LimitsError.credentialMissingAccountOAuth` and showing a misleading "re-login required" banner that could never be resolved by `claude login`.

Claude Code CLI itself queries and writes using `-a <macOS username>`:
```bash
security find-generic-password -a <macOS username> -w -s "Claude Code-credentials"
```

### Changes
- Added `kSecAttrAccount as String: NSUserName()` to the Keychain query dictionary via `OAuthCredentialData.claudeKeychainQuery`.
- Added unit tests in `PokeTokenBarTests.swift` verifying query attributes, account scoping, and no-UI flag behavior.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## Checklist

### DIFF
--- a/Sources/PokeTokenBar/Core/OAuthLimitsProvider.swift
+++ b/Sources/PokeTokenBar/Core/OAuthLimitsProvider.swift
@@ -237,15 +237,8 @@ private actor OAuthAccessTokenCache {
         if KeychainAccessGate.isDisabled {
             throw LimitsError.keychainAccessDisabled
         }
-        var query: [String: Any] = [
-            kSecClass as String: kSecClassGenericPassword,
-            kSecAttrService as String: OAuthCredentialData.claudeKeychainService,
-            kSecReturnData as String: true,
-            kSecMatchLimit as String: kSecMatchLimitOne,
-        ]
-        if !allowKeychainPrompt {
-            KeychainNoUIQuery.apply(to: &query)
-        }
+        let query = OAuthCredentialData.claudeKeychainQuery(
+            allowKeychainPrompt: allowKeychainPrompt)
 
         var item: CFTypeRef?
         let status = KeychainReader.copyMatching(query, &item)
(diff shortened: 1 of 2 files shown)


justinjeong5:
Hi @rel94! Thanks for investigating and putting up a fix for this.

I ran into the exact same issue and opened #232 around the same time. Scoping by `NSUserName()` in #231 is very clean.

One small edge case to consider: depending on how Claude Code or credential helpers write to the keychain (e.g. across `CLAUDE_CONFIG_DIR` or custom setups), the account name attribute might not always strictly match `NSUserName()`. In #232, I used `kSecMatchLimitAll` to iterate through entries and pick whichever one successfully parses valid `claudeAiOauth` credentials.

Whichever approach @chattymin prefers (or if you'd like to incorporate the dynamic scanning into this PR), glad we got the root cause addressed!

---

chattymin:
Thanks for reporting this in #229 and following up with a fix, @rel94 — the diagnosis in that issue is what made the problem actionable, and pulling the query into `claudeKeychainQuery` was a clean refactor.

I merged #232 instead. Both PRs fix the same bug; they differ in what they assume about the Keychain item:

- **#231** scopes the query with `kSecAttrAccount: NSUserName()`.
- **#232** uses `kSecMatchLimitAll` and picks whichever returned item parses into a valid `claudeAiOauth` credential.

@justinjeong5 raised the deciding point in the comment above: the account attribute is not guaranteed to equal `NSUserName()`. Credential helpers and custom setups can write a different `acct`, and on those installs #231 would turn a lookup that works today into one that fails. I went with the approach that does not depend on the account naming convention.

So this is a design call between two valid fixes, not a problem with your work — and the fix that landed exists because you reported the issue. Thank you.

PR #235, opened 2026-08-28
feat: auto-enable floating pet when a representative is pinned from the dex

aOvOb:
## Summary

Picking a representative Pokémon in the dex previously had no visible effect if the floating pet toggle in Settings was off, since the two UIs don't know about each other. 
CompanionStore now exposes an `onRepresentativeSet` callback, fired only when a species is actually pinned (not when clearing back to auto-follow), which AppDelegate wires to flip `UsageStore.floatingPetEnabled` on. 
so pinning a Pokémon from the dex makes it show up on the desktop immediately.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes
No UI files changed

## Checklist

### DIFF
--- a/Sources/PokeTokenBar/Core/CompanionStore.swift
+++ b/Sources/PokeTokenBar/Core/CompanionStore.swift
@@ -44,6 +44,8 @@ final class CompanionStore {
     private let dittoDisguiseRollingEnabled: Bool
     /// 세션 내 활성 개체 교체 감지용. await 뒤 이전 개체의 결과가 새 개체를 덮지 않게 한다.
     private var activeGeneration = 0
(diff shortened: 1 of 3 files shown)


chattymin:
Thanks for the idea, @aOvOb — and the implementation is clean, especially routing it through a callback instead of reaching into the store from the dex view.

I'm going to pass on this one though. Pinning a representative already shows up where it needs to: the menu bar follows the pinned species, which is the surface the feature was built for. The floating pet is a separate surface with its own explicit on/off setting, so the two are not preconditions for each other — pinning a representative does not imply wanting a pet window on screen, and someone who deliberately turned the pet off would see it reappear without asking for it.

Thanks again for taking the time.

Issue #236, opened 2026-08-28
Kiro usage is not detected after session storage moved from SQLite to JSONL


TeiNam:
Environment

PokeTokenBar: 2.5.2

Kiro CLI: 2.20.1

Kiro CLI v3 enabled via kiro-cli --v3

Problem

PokeTokenBar currently reads Kiro usage from:

~/Library/Application Support/kiro-cli/data.sqlite3

The database still exists, but recent Kiro sessions are no longer written there. As a result, PokeTokenBar reports no Kiro usage.

Current storage locations

Kiro CLI 2.20.1 stores sessions at:

~/.kiro/sessions/cli/*.jsonl

Kiro CLI v3 stores sessions at:

~/.kiro/sessions/\<workspace\>/\<session\>/messages.jsonl
~/.kiro/sessions/\<workspace\>/\<session\>/session.json

Adding ~/.kiro as an additional scan folder does not help because the current Kiro provider expects a data.sqlite3 file inside each configured root.

Expected behavior

PokeTokenBar should detect recent Kiro usage from the new JSONL session formats while retaining the legacy SQLite reader for older installations.

For v3 sessions, messages.jsonl also contains usage_summary events with credit usage. These credits should either be reported separately or token usage should continue to be clearly marked as an estimate.

Borisserz:
Taking this.

Kiro CLI 2.20+ and `--v3` stopped writing new sessions to `data.sqlite3`. The current reader only looks for that database under each root, so extra-folder `~/.kiro` is a no-op and recent usage shows 0.

Plan: keep the SQLite path for older installs, and also scan the JSONL layouts (`~/.kiro/sessions/cli/*.jsonl` and `~/.kiro/sessions/<workspace>/<session>/messages.jsonl`). Tokens stay a bytes÷4 estimate (`reportsCost = false`) — credits in `usage_summary` are not dollar cost.

chattymin:
Fixed in #238, which shipped just now.

Both layouts you documented are read, alongside the legacy store:

- `~/.kiro/sessions/cli/*.jsonl` (2.20.x)
- `~/.kiro/sessions/<workspace>/<session>/messages.jsonl` + `session.json` (v3)
- `~/Library/Application Support/kiro-cli/data.sqlite3` stays as the fallback for older installs

[...]

PR #238, opened 2026-08-28
fix: read Kiro CLI 2.20+ JSONL sessions (#236)

Borisserz:
## Summary
- Kiro CLI 2.20+ / `--v3` stopped writing new sessions to `data.sqlite3`, so usage showed 0 even when `~/.kiro` was added as an extra scan folder (the reader only opened sqlite under each root).
- Scan the writer-shaped JSONL layouts (`sessions/cli/*.jsonl` and `sessions/<workspace>/<session>/messages.jsonl`) while keeping the legacy SQLite path. Tokens stay a bytes÷4 estimate; `usage_summary` credits are not converted to USD (`reportsCost` stays false).
- Register `KIRO_HOME` / `KIRO_CLI_HOME` (and the same-class miss `CURSOR_DATA_DIR`) in `UsageEnvironment.names` so GUI apps can see those overrides.

## Test plan
- [x] Fixtures copied from tokscale/codeburn Kiro parser tests (Prompt/AssistantMessage envelope, structured `payload.type`, flat `{role,content}`)
- [x] `testExtraRootFindsJsonlSessionsWithoutSqlite` — the #236 extra-folder no-op
- [x] SQLite + JSONL both counted; unrelated `*.jsonl` ignored
- [x] Inject-fail: disabling the JSONL scan in `kiroEntries` turns the new tests red
- [x] v3 flat multi-turn without timestamps does not collapse under `dedupKeepMax`
- [x] `./scripts/test-gate.sh` — 796 tests, 0 failures, logic-core 90.46%


chattymin:

Issue #240, opened 2026-08-30
Optional claude.ai session key: official limits without a Keychain prompt


LuceteYang:
## Problem

Two rules in `docs/reference/defect-log.md` ("자격증명·Keychain") corner official Claude limits:

1. **Auto-polling must never read the Claude Keychain** — a no-UI query suppresses the *authentication* prompt but not the *unlock / allow-access* dialog. Measured there: `SecItemCopyMatching` blocking 13s mid-poll with a dialog, a few times a day around token expiry.
2. **The app must not create its own Keychain item** — an app-owned item's ACL breaks on every code-signature change, and that prompt can't be suppressed (#58).

So once the cached token expires and `~/.claude/.credentials.json` is absent, limits freeze until the user presses **Refresh limit token cache** — a manual step that lands on exactly the users already seeing prompts.

The same entry names the only complete cure and marks it **not adopted**: Developer ID notarization, because an ACL grant only persists for a trusted signing identity. That's a paid path and a reasonable one to skip. This is for the users hitting the prompt loop meanwhile.

**As a user, before I had an alternative:** every Keychain access raised a password dialog, and "Always Allow" didn't hold — it returned a few times a day, and again after every app update, because the new cdhash invalidated the grant.

## Proposal

An **opt-in** second source: a claude.ai `sessionKey` cookie pasted under **Settings → Advanced**.

- Same 5-hour / weekly numbers from `claude.ai/api/organizations/{org}/usage` — same schema as `api.anthropic.com/api/oauth/usage`, so `LimitStatus` is reused unchanged and popover rows don't change shape.
- Touches no Keychain, so **auto-polling can use it** — limits stay current, and there's no ACL grant to lose on the next update.
- No key saved → nothing changes and no file is created. Chain order: session key first when present, otherwise today's OAuth path.

## Prior art

Several macOS/VS Code trackers already use this exact endpoint and credential — [linuxlewis/claude-usage](https://github.com/linuxlewis/claude-usage) (has a written SPEC), [hamed-elfayome/Claude-Usage-Tracker](https://github.com/hamed-elfayome/Claude-Usage-Tracker), [f-is-h/Usage4Claude](https://github.com/f-is-h/usage4claude) (also surfaces 7-day Opus/Sonnet), [DimaNeskorodiev/claude-limits](https://github.com/DimaNeskorodiev/claude-limits). The approach and its local-storage model are established, not novel.

## The tradeoff, up front

[...]

PR #241, opened 2026-08-30
feat: fetch official limits with a claude.ai session key (no Keychain prompt)

LuceteYang:
## Summary

Implements the proposal in #240 — an **opt-in** claude.ai `sessionKey` path for official Claude limits, so they stay current without a Keychain prompt.

Auto-polling never reads the Keychain by design (defect log, "자격증명·Keychain"). So once the cached token expires and `~/.claude/.credentials.json` is absent, limits freeze until the user presses **Refresh limit token cache** — and on an unnotarised build that press raises the password dialog, because an "Always Allow" grant does not survive the next code-signature change.

With a key saved, limits come from `claude.ai/api/organizations/{org}/usage`, which returns the same schema as `api.anthropic.com/api/oauth/usage` — `LimitStatus` is reused unchanged, so no popover row changes shape. **Without a key nothing changes:** `ChainedLimitsProvider` falls through to today's OAuth path and no file is created.

**Opening this so the shape is concrete rather than hypothetical.** The open questions in #240 still stand — most importantly whether plaintext file storage is acceptable at all. If it isn't, please close this; the answer is worth more than the code. I'm equally happy to cut it into a smaller first slice (e.g. drop the organization picker).

Three details worth reviewing specifically:

- **Organization selection is not "the first one".** A claude.ai account also returns orgs the user merely logged into, and API-only orgs that answer 403 — picking first yields a permanent, plausible-looking 0%. Candidates are filtered to chat-capable orgs, probed in parallel, and the one with usage is cached; a cached org that later 403s triggers rediscovery.
- **A dead key answers 403, not 401.** Verified by flipping one character of a stored key in a running build: the usage endpoint *and* `/api/organizations` both returned 403. Since `/api/organizations` is not org-scoped, a 403 there means the key itself can see nothing, so it maps to `sessionKeyInvalid`. Per-org 403s keep their meaning and still trigger rediscovery.
- **Expiry carries its source** (`limitsAuthExpiry: .oauth | .sessionKey`) instead of one boolean, because the remedies are opposites. Collapsed into one flag, a session-key user is told to run Claude Code once — useless for a dead cookie — and the retry button reads the Keychain, raising the prompt the key exists to avoid.

**Storage:** owner-only (`0600`) plaintext file, not the Keychain. An app-owned Keychain item is the very class of defect this removes — a cdhash change invalidates its ACL and the prompts return. The settings copy states the storage model and that a browser logout invalidates the key immediately.

> **Testing note:** `curl` cannot verify these endpoints. Cloudflare answers it with its own 403 challenge regardless of cookie, so it will confirm any theory you bring it. Only `URLSession` gives a real answer; the unit tests seam out the HTTP layer.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

All new UI lives inside the existing **Settings → Advanced** disclosure, plus one conditional banner in the popover's limits section.

pr-session-key-ui

| Before | After |
| ------ | ----- |
| Settings → Advanced starts with "Disable Keychain access". | A **claude.ai session key** row precedes it: label with a status badge, a hint, a `SecureField` (never renders the saved value back — placeholder only), **Save** / **Delete**, an organization picker when more than one candidate is visible, and a note stating the key is stored unencrypted and that a browser logout invalidates it. |
| Badge shows nothing, or **Saved** in green once a key exists. | Additionally shows **Expired** in orange while a stored key is being rejected, so a user who followed the expiry banner into Settings can see what to act on. |
| An expired Claude credential shows one banner: "Claude session expired", retry button reads the Keychain, hint says to run Claude Code once. | With a session key, that case instead shows "claude.ai session key expired" with a **Settings** button that opens Settings with Advanced already expanded (the key field lives inside a collapsed disclosure, so opening Settings alone would be a dead end). The OAuth banner is unchanged for users without a key. |

## Checklist

`swift test`: 881 tests, 0 failures on this branch.


chattymin:

PR #246, opened 2026-08-31
feat: explain delayed egg hatches and harden retry state

justinjeong5:
## Summary

Explains why a fully incubated egg has not hatched yet when its PokéAPI data cannot be resolved, while keeping the egg and retrying on the next refresh.

This complements the existing reconnect refresh: reconnection triggers another refresh, while this change gives the user accurate feedback during the delay.

## Changes

- Track a transient `isHatchRetryDelayed` state only for the current ready egg.
- Prefer hatching over prefetching once the egg reaches its difficulty-scaled threshold.
- Guard every asynchronous failure result with the current generation and egg state so stale requests cannot affect an imported save or replacement egg.
- Carry prefetch failures into the ready-egg delay state and continue hatching after a successful in-flight prefetch.
- Show a cause-neutral “retrying on the next refresh” message in all 7 supported languages.
- Clear the delay state after recovery, hatching, egg replacement, graduation, or save import.

## Tests

- Delay state after a failed hatch and reset after provider recovery.
- Production `update`/prefetch interleaving when the egg becomes ready during an in-flight request.
- Stale species-selection and line-fetch failures after save import.
- Save-import reset behavior.
- Exact localized copy for all 7 supported languages.

## Verification

- GitHub CI `build-test` — 1,083 tests executed, 11 skipped, 0 failures.
- Logic-core line coverage — 92.94% (required: 75%).
- GitHub CI `secret-scan` — passed.
- Local `swift build`, test-source parse, and `git diff --check` — passed.
- 3-agent adversarial review across logic, compatibility, security, performance, design, and tests — no remaining blocker or major findings.
- Quality gate — 4.62/5 (`> 4.5`).

---


chattymin:

PR #251, opened 2026-08-31
feat(kun): add Kun CLI usage provider

Legiani:
## Summary

Adds Kun CLI (`kun`) as a 13th local usage source. Kun persists session usage in an SQLite database under `~/.kun/data/index.sqlite3` in the `usage_events` table with `(thread_id, seq, timestamp, model, usage_json)`.

Key details:
- **Per-turn delta computation**: Kun records cumulative session stats in `usage_json` (`promptTokens`, `completionTokens`, `cachedTokens`, `totalTokens`, `costUsd`); the reader calculates per-turn token and cost deltas across sequential events in each thread.
- **Durable roots & environment**: Default root is `~/.kun/data/index.sqlite3` with `KUN_DATA_DIR` override support registered in `UsageEnvironment`, `CustomScanRoots`, and tests.
- **Extension contract compliance**: Implements `UsageProvider` and registers in `UsageStore.init` `providers:` array without provider-specific logic in generic paths.

## Type of change

- [x] New feature

## UI changes

No UI structure changes. Users with a Kun database get a `Kun` capsule in the provider tab bar and an entry in Settings → Additional scan folders; Kun token usage and cost join today's aggregate totals.

## Checklist

- [x] `swift build` and `swift test` pass locally (907 tests, 0 failures)
- [x] Tests added in `KunUsageTests.swift` and registry tests updated

### DIFF
--- a/Sources/PokeTokenBar/Core/CustomScanRoots.swift
+++ b/Sources/PokeTokenBar/Core/CustomScanRoots.swift
@@ -104,6 +104,8 @@ enum CustomScanRoots {
         case "omp":
             return CustomScanRoots.union(
                 defaults: LocalUsageReader.computeOmpSessionRoots(), extraRaw: nil)
+        case "kun":
+            return LocalAdditionalUsageReader.kunRoots(customRootsValue: nil)
         default:
             return []
         }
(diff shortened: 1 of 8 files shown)


chattymin (changes requested):
Please preserve already observed usage when Kun threads disappear. Merge cache entries by a stable entry ID, keep the larger cumulative value for an existing thread, and retain previously observed entries that are absent from a later scan instead of replacing the cache wholesale. Please add a regression test proving that deleting a thread cannot reduce historical totals, update the data-source documentation, and rebase onto current main.

PR #254, opened 2026-09-01
feat: grant 2x growth to repeat hatches

### LINKED ISSUE #253: Repeat hatch of an already-graduated line costs the full progression again
## What happened

My 8th egg hatched Ducklett — the very first line I graduated three weeks ago. Nothing about it is new: not shiny, a linear line so no new final form to collect, and nature is already changeable with a Mint. It still needs the full 750M to graduate, exactly like a brand-new Common.

## Why it feels wrong

Repeats are intentionally allowed: [`chooseBase`](https://github.com/chattymin/PokeTokenBar/blob/37763d3c367068492c18f6e51b45977c2d27f6d5/Sources/PokeTokenBar/Core/CompanionStore.swift#L1092-L1094) halves the weight of already-collected bases to keep re-hatch and shiny hunting open, and [`pickPlannedChild`](https://github.com/chattymin/PokeTokenBar/blob/37763d3c367068492c18f6e51b45977c2d27f6d5/Sources/PokeTokenBar/Core/CompanionStore.swift#L527-L532) already steers branching lines toward uncollected finals. I'm not asking to remove any of that.

The gap is on the cost side. A repeat of a linear line gives no Pokédex progress (unless it happens to be shiny), but it is priced like a new catch — [the full graduation total](https://github.com/chattymin/PokeTokenBar/blob/37763d3c367068492c18f6e51b45977c2d27f6d5/Sources/PokeTokenBar/Core/CompanionModel.swift#L113-L120) for its rarity. So the egg reveal, which is the best moment in the loop, becomes "nothing new, and you're committed to the full cycle anyway."

For scale, I simulated the 328-species pool with capture-rate weights and the ½ penalty (method below): the median user meets their first repeat around the 25th graduation; roughly 5% of hatches are repeats by #20 and 14% by #60. So this is rare and a change here has a small blast radius — but when it lands, it is several days of the full cycle for nothing, and weeks for the lighter users described in #226.

## Suggestion

When the hatched base species already has an entry in `collectedFinals` — the same predicate `chooseBase` uses for the weight penalty — automatically halve that Pokémon's evolution and graduation thresholds. Surface it in Home as "2× growth" with the adjusted remaining-token count. No prompt and no item, so the "you code, something grows" loop stays exactly as it is.

Two details that I think matter:

- Key the discount on the **base species**, not the planned final form. The planned path is deliberately not shown in the UI, and a discount tied to the final form would leak the branch choice through the remaining-token number.
- Decide the discount at hatch and persist it on the active Pokémon, so it survives restarts and save import/export.

A cheaper alternative (or complement) is to lower the collected-base weight further than ½ — e.g. 1/10 — which is a one-line change in `chooseBase`. Either direction works for me; the exact multiplier is a balance call.

## Implementation note

With halved thresholds the smallest phase threshold becomes 62.5M (3-form Common, stage 0). One Rare Candy (100M) still cannot chain two stages, because the carry-over (< 100M) is below the next threshold (125M), so the "one candy, one stage" invariant in [`RareCandy.xp`](https://github.com/chattymin/PokeTokenBar/blob/37763d3c367068492c18f6e51b45977c2d27f6d5/Sources/PokeTokenBar/Core/CompanionModel.swift#L173-L177) still holds. Worth pinning with a test if this goes in.

Distinct from #226, which is about the global token-to-EXP pace for everyone; this is only about the relative cost of a Pokémon you have already finished.

<details>
<summary>Simulation method</summary>

Monte Carlo over the cached base index (`~/Library/Application Support/PokeTokenBar/base-index.json`, 328 entries), reproducing `chooseBase`'s weighting. 4,000 runs of 60 consecutive hatches each; every hatch is added to the collected set.

```python
import json, random
e = json.load(open("base-index.json"))["entries"]
ids, cr = [x["id"] for x in e], [x["captureRate"] for x in e]

def run(n=60):
    col, hits = set(), []
    for _ in range(n):
        w = [max(1, c // 2) if i in col else c for i, c in zip(ids, cr)]
        b = random.choices(ids, weights=w)[0]
        hits.append(b in col); col.add(b)
    return hits
```

`P(hatch #k is a repeat)` is the mean of `hits[k-1]` across runs; "first repeat" is the index of the first `True`. Guaranteed-tier eggs narrow the pool and push these numbers up slightly.
</details>

parkjun5:
## Summary

- Grant a persisted 2× growth boost when the hatched base species already has a graduated entry.
- Route progress, evolution, graduation, and Ditto reveal thresholds through the active Pokémon growth multiplier.
- Show a localized 2× growth badge and the adjusted remaining-token count in the companion header.
- Cover branching paths, legacy saves, save transfer, Rare Candy, Ditto reveal, and localization with regression tests.

Closes #253

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Repeat hatches use the standard progression cost and have no distinct growth indicator. | Repeat hatches use halved phase thresholds and show a localized 2× growth badge beside the stage and nature. |

## Testing

- `swift build` passes locally.
- `swift test` could not compile in the local environment because only Command Line Tools are installed and the selected toolchain does not provide `XCTest`; CI must confirm the full suite.

## Checklist

- [ ] `swift build` and `swift test` pass locally


chattymin:

PR #258, opened 2026-09-02
Turn charms into an unbounded tiered ladder

donky-ey:
Charms were one-shot purchases: pay once, get a fixed effect forever. Each is now a ladder that can be climbed indefinitely.

## The ladder

- A tier costs double the one below it (`250M x 2^(t-1)`) while the effect grows by a fixed step, so the cost per unit of benefit doubles every step. There is no cap because the price is the cap.
- Tier 4 reproduces the old effect exactly - 2.00x experience, 1.50x currency, 1/48 shiny - and that is where existing owners are migrated at decode time. The legacy booleans are kept rather than deleted, so downgrading still sees the purchase.
- The shiny denominator has a floor of 1/16. The Rainbow Charm now takes a flat -8 at any tier instead of pinning the odds to 1/32, so it keeps meaning however far the Shiny Charm is climbed.

## Display

Every charm row names what it raises, what it does now, and what the next tier would do - "EXP 2.00x -> Tier 5 2.25x" - with the shiny charm speaking in denominators, the notation players know. The Bag says the same thing beside each charm, and both screens build the string from one place in `L` so they cannot drift apart.

## Defects found while building this

- **The migration never ran.** `ownsShinyCharm` and `ownsRainbowCharm` were decoded *after* the migration block read them, so it always saw `false` and silently did nothing. All four legacy booleans now decode ahead of it.
- **`maxSafeTier = 40` was not safe.** The cumulative price overflows `Int` at 36, and Swift's trapping multiply killed the process - the test run died with SIGTRAP. The bound is now derived from `Int.max` rather than hardcoded.
- **The Rainbow Charm regressed.** A flat -8 gives 1/56 where the old charm pinned 1/32, which would have made completing the National Dex worth less than before. It now carries 7 shiny tiers as well, which lands on exactly 32.

Every guard above is covered by a regression test, and each was mutation-checked: reverting the guard fails the test that names it.


donky-ey:
Opened against the wrong repository; reopening on donky-ey/PokeDexBar.

PR #264, opened 2026-09-03
feat: give each Pokémon persistent individual values and a Pokédex detail page

CCedDev:
## Summary

Every hatched Pokémon now has **individual values that persist**, so the same creature stays the same creature across launches: IVs, gender, ability, and a level-up moveset that follows its growth. Tapping a Pokédex cell opens a **detail page** with nature-modified computed stats, species data, and the full supported learnset.

A Pokédex species can aggregate several catches, so the page has an individual picker. Earlier evolution stages stay species reference pages rather than being duplicated as separate creatures.

### What it is for

Right now the companion is described by very little: species, stage, rarity, shiny, nature. That is enough to draw a sprite and a progress bar, but it does not describe a *creature* — two Charizards raised on the same token budget are indistinguishable. This PR gives each one an identity that survives evolution, restart, and save transfer, and then surfaces it in a place users already visit.

### Where this is heading (deliberately not in this PR)

This is groundwork for a possible **local-network Pokémon battle** between two PokeTokenBar users — you raise a companion on your own token usage, meet someone on the same Wi-Fi, and let the two collide. No server, no accounts, no matchmaking; strictly peer-to-peer on the LAN, in line with the project's on-device-first stance.

This PR contains **no battle code and no networking**. What it does is make the missing half exist:

- A battle needs exactly what a main-series damage formula needs — base stats, IVs, level, nature modifier, ability, and up to four moves. `PokemonStatCalculator` already produces main-series stat values from that set, and `PokemonProfile` is the persisted half.
- `PokemonProfile.instanceID` gives every creature a stable identity, which is what lets you name a combatant across two devices instead of just "a Charizard".
- The existing `SaveTransfer` envelope is the pattern a peer exchange would reuse: schema-versioned, header-checked, and `sanitized()` at the boundary — it already treats the other side's bytes as untrusted, which is precisely the posture you want when the other side is another machine on the network. Bumping it to v2 and clamping profiles on import (below) is part of that.

I have kept this PR to the local, useful half on purpose: the profiles and detail page stand on their own as a feature, and none of it commits the project to shipping battles. If you would rather this went a different direction, the profile model is the only piece a battle would depend on, so it is cheap to redirect.

### Data handling

Per [CONTRIBUTING](https://github.com/chattymin/PokeTokenBar/blob/main/CONTRIBUTING.md), no assets are committed. Stats, abilities, and moves are fetched at runtime from PokéAPI and cached only under `~/Library/Application Support/PokeTokenBar/`. The parser keeps only the supported version group (`black-2-white-2`, matching the Gen-V animated sprite ceiling) **before** the response reaches memory or disk — 82 KB → 5.5 KB per species for Pikachu, measured against `pokeapi.co/api/v2/pokemon/25`. Only the rolled individual values live in the companion save.

### Three things worth a reviewer's attention

- **Save schema 1 → 2.** v1 saves still import (the gate is `<=`), but a v2 save cannot be imported by an older build. A one-time, never-overwritten `companion-state.pre-profiles-v1.json` backup is written before migrating, and imported profiles are clamped at the trust boundary in `SaveTransfer.sanitized`.
- **Ditto stays honest.** Revealing a disguise rebases the profile, so a revealed Ditto reports Ditto's genderless identity, ability, and moves instead of the disguise species'. Reveal and path normalization both kick a detail load, matching what the evolution branch already did.
- **`scripts/test-gate.sh` now covers `PokemonProfile.swift`.** This tightens what CI enforces.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Tapping a Pokédex cell selected it; the footer line showed `#id name · rarity` plus the "set as representative" button. | Tapping a cell opens a scrollable detail page: sprite + name + rarity header (with the representative button moved here), an individual picker when a species has several catches, an "Individual" card (level, gender, nature, ability, computed stats with an IV column, known moves), a "Species data" card (types, height, weight, base-stat total, possible abilities), and the complete move list. A back button returns to the grid, which is otherwise unchanged. |
| — | While details load the page shows a spinner; if the fetch fails it shows a message plus a retry button. |

Stat bars scale to the highest displayed value rounded up to the next 100 (floor 300), so they stay readable at level 100 instead of pegging. New localized strings are in all seven supported languages. The canonical `assets/` screenshots are regenerated at release, not in this PR.

image

## Checklist

- [x] No copyrighted assets, secrets, or private tooling references are committed (see [CONTRIBUTING](https://github.com/chattymin/PokeTokenBar/blob/main/CONTRIBUTING.md))

Local verification: `swift build` clean, **929 tests / 0 failures**, `./scripts/test-gate.sh` green at 91.69% line coverage. `PokemonProfile.swift` and the new `CompanionStore` block have **no unexecuted regions** under `llvm-cov --show-regions`, not just no unexecuted lines. The new `sanitize()` clamps were verified by injecting the defect: disabling either `SaveTransfer` call site fails its own test, and only its own test.

https://claude.ai/code/session_01SyRfL85nwsaUytvcNA9qqm


janincloud:
_Jans Claude here_ 👋

Reviewed the full diff. Solid, unusually well-tested change — two new test files (~530 lines) covering migration, sanitization, RNG determinism, offline fallback, and lock lifetime. CI is green and there's nothing outstanding from other reviewers. A few honest notes, none blocking.

### Should note (design constraint, currently unreachable)
`normalizedMoves` filters everything to the `black-2-white-2` version group. Any species not present in that learnset yields `moves == []`, so both the individual moveset and "Complete move list · 0" render empty. This is safe **only** because species are already bounded to Gen V by animated-sprite availability. If that constraint ever loosens (Gen VI+ sprites), detail pages silently go move-less. A one-line comment tying the version-group choice to the sprite/species bound would keep that invariant from rotting.

### Nits (non-blocking)
- `PokemonProfile.applyGrowth` floors level at **5**, but `sanitize()` floors it at **1**. Only reachable via a hand-edited imported save, so harmless — but the two floors could agree.
- `CompanionStore.applyUsage` runs the full `enrich` (ability fallback resolution + `levelUpMoves` recompute) on every token delta, including `delta == 0` calls (reveal, line-load). It's idempotent and cheap, so purely cosmetic — just slightly more work than needed on no-op ticks.

### Verified as correct (not issues)
- `enrich` is idempotent for identity fields (gender/ability locked once set) while intentionally letting `moves` grow with level — the RNG draw ordering holds because the gender draw is only consumed when `gender == nil`.
- Ability slot + hidden flag preserved across evolution and re-resolved by slot against the new species — matches real Pokémon mechanics.
- `loadPokemonDetails` dedup guard has no TOCTOU (check-then-insert is synchronous before the `await`); the "already loading" early return is safe because the view reads `pokemonDetailsByID` reactively and `enrichProfiles` fires on completion.
- Migration is additive, network-free, idempotent, and takes a one-time `pre-profiles-v1` backup before the first format expansion. `SaveTransfer.sanitized` correctly clamps profiles at the import trust boundary (the `moves` clamp matters — that list renders eagerly in a non-lazy `VStack`).

### Verdict
Ship-ready. Clean schema-v2 migration with a safety backup, a proper trust boundary on imported saves, deterministic offline-friendly enrichment, and thorough tests. The only thing worth a follow-up is documenting *why* `black-2-white-2` is the correct learnset cutoff, since that silently depends on the Gen-V sprite constraint.

chattymin:

PR #267, opened 2026-09-04
feat: raise the floating pet size ceiling from 192px to 384px

yojkim-lbox:
## Summary

The desktop floating pet's size slider topped out at 192px. This doubles the
maximum to 384px so the companion can be made noticeably bigger on the desktop.

Nothing else about the pet changes — the minimum (48px) and the 8px step stay as
they are, and no stored value can exceed the new maximum, so there is no
migration concern.

Two details that make the larger sizes safe:

- `SpriteView` draws with `.interpolation(.none)`, so a 96×96 sprite scaled to
  384pt stays crisp pixel art rather than blurring.
- `FloatingPetController.observeSettings()` already tracks `store.floatingPetSize`,
  so dragging the slider resizes the `NSPanel` and the sprite together — no clipping
  at the new range.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Settings → Floating pet → Size slider ranges 48–192px. At the maximum the pet is 192pt on a side. | Same slider, same 8px step, now ranging 48–384px. Existing users see no change until they move the slider; their stored size is unaffected. |

The three READMEs (en/ko/ja) documented the "48 to 192px" range in the floating-pet
section and are updated to match. The canonical `assets/` screenshots are unchanged,
since they are regenerated at release.

## Checklist

- [x] `swift build` and `swift test` pass locally (905 tests, 0 failures)

### On the test

Raising the ceiling makes a previously rare branch the common case:
`panelSize` returns `max(petSize, bubbleMinWidth)`, and with `bubbleMinWidth = 180`
only the last two slider stops (184, 192) let the pet drive the panel width. It had
no coverage. `testPanelGrowsForBubbleWithoutChangingPetOrigin` now also asserts the
panel geometry at the new maximum. Verified by fault injection — hardcoding
`bubbleMinWidth` as the panel width fails the new assertions.

https://claude.ai/code/session_01X91ErNfhphSw9gC1pqjjjN


chattymin:

Issue #269, opened 2026-09-05
Feature: this month's day-by-day usage trend in the popover


LuceteYang:
Week and month usage each collapse to a single `PeriodUsage` scalar (`Core/Models.swift`), so the popover can say **how much** this month but not **when**. No way to spot a spike, a quiet stretch, or whether today is unusual.

The data is already there and thrown away. `enrichmentScanStart` covers the whole current month for every provider, `Entry` already carries `localDay`, and `period()` filters on it, keeps the sum and drops the rest. A daily series is a group-by over entries that are already in memory — no `Entry` field, no cache version bump, no re-parse, no new I/O, no change to the scan window.

**Proposal:** an optional `monthDaily` on `ProviderEnrichment`/`ProviderSnapshot`, filled by one shared assembly helper so it stays provider-agnostic, and a compact bar row inside the existing popover.

![Daily usage for the current month](https://raw.githubusercontent.com/LuceteYang/PikaTokenBar/feat/monthly-daily-trend/assets/screenshot-daily-trend.png)

**Limits, up front:**

1. **Current month only.** No rolling 30 days, no previous month — the scan window already covers exactly this range.
2. **If the logs are gone, that day reads as 0 and cannot be recovered.** This is a view of the logs, not a ledger.
3. **Cross-month sessions are deliberately truncated.** A session that began last month and ran into this one has a current-month mtime, so it is read in full and last month's entries come along. Drawing those would show a partially filled previous month, since the other files from last month were never scanned.

#260 (daily streak) needs per-day data, which does not exist today; this would be that infrastructure.

I implemented it in a fork and have been running it for a day — PR opened for reference. **Happy to close it if the shape isn't what you want**; I'd rather agree on that first.

PR #270, opened 2026-09-05
feat: show this month's day-by-day usage trend in the popover

LuceteYang:
Implements #269. **Reference PR — close it freely if the shape isn't what you want.**

Daily usage for the current month

A bar per day of the current month under the week/month totals, with a date axis (day 1, every 7th, always today), a tick under weekends, and a caption naming the day you point at — today when you point at nothing.

**Cost:** none. `enrichmentScanStart` already loads the whole current month on every poll and `period()` throws everything but the sum away. No `Entry` field, no cache version bump, no re-parse, no new I/O, no scan-window change.

**Two things worth reviewing:**

- `monthDailySeries` builds the date axis **from the month range** instead of grouping entries by `localDay`. The scan is an mtime filter, so a session that began last month and ran into this one drags last month's entries in; emitting them would draw a partially filled previous month. Unused days are explicit zeros — bar position is the date.
- The twelve copied `fetchEnrichment` bodies now share `ProviderEnrichment.local(entries:now:zeroCost:)`. Net −126 lines, no behaviour change. If you'd rather keep them separate, say so and I'll add `monthDaily` to each copy instead.

`monthDaily` is optional — a provider that can't produce a series stays `nil` and drops out of the aggregate.

**Tests** (18 new, suite 924 green, `test-gate.sh` at 91.6%): cross-month truncation through a real mtime scan, dense zeros, `sum(series) == monthTotal`, providers reporting no series, axis label collisions across all 31 possible "today" values, caption staying one line in all 7 languages. Time zone and calendar are injectable — a fixed 86400-second day stride passed the axis test on a machine without DST and only failed under `TZ=America/Los_Angeles`.

README/landing updates are ready if you want the feature; left out here to keep the diff reviewable.


chattymin:

PR #276, opened 2026-09-07
feat: break omp usage down per model in the popover

yourconscience:
## Summary

`omp` (oh-my-pi) has its own reader and provider (`~/.omp/agent/sessions`, `LocalOmpProvider`/`parseOmpLine`), separate from the Pi path that #225 gave a per-model breakdown. This brings `omp` to the same treatment, mirroring `LocalPiProvider`:

- **Per-model daily breakdown.** `LocalOmpProvider.fetchDaily` now calls `daily(includeModels:)`, so `DailyUsage.models` is populated and the popover's existing generic per-model rows render for `omp` too.
- **Injectable cache seam.** `init(cache:)` is added to `LocalOmpProvider`, mirroring `LocalPiProvider`, so `fetchDaily` can be exercised end to end.

Deliberately out of scope, per review:

- **No parser change.** Model attribution stays as-is: it comes from `message.model`, which the official OMP session writer records. No envelope-level override, so no cache-version migration.
- **Cost is left to the parser.** `omp` records its own model-price estimate in `usage.cost.total`. #289's parser preserves that recorded amount and surfaces it as an estimate (`costCoverage == .estimate`), not a billed charge, and `fetchDaily` passes it through untouched.

Rebased onto current `main` (`09fd600`); #289's cost handling and the shared enrichment path are preserved.

## Test plan

- New provider-level test `testProviderFetchDailyForwardsBreakdownAndKeepsSourceCost`: routes a fixture through `LocalOmpProvider.fetchDaily` and asserts the per-model breakdown survives provider repackaging and the source-recorded cost is preserved as an estimate.
- CI (macos-15): full suite 1078 tests, 11 skipped, 0 failures; logic-core line coverage 92.94% (gate 75%).
- `swift build` clean locally.

### DIFF
--- a/Sources/PokeTokenBar/Core/LocalUsageProvider.swift
+++ b/Sources/PokeTokenBar/Core/LocalUsageProvider.swift
@@ -131,16 +131,23 @@ struct LocalPiProvider: UsageProvider {
 struct LocalOmpProvider: UsageProvider {
     let id = "omp"
     let displayName = "omp"
+    /// 캐시 시임 — 기본은 공용 캐시(실 로그), 테스트만 픽스처 루트를 주입한다.
+    let cache: LocalUsageCache
+
+    init(cache: LocalUsageCache = .shared) { self.cache = cache }
 
     func fetchDaily() async throws -> DailyUsage? {
         let now = Date()
-        let entries = await LocalUsageCache.shared.ompEntries(modifiedSince: Calendar.current.startOfDay(for: now))
-        return LocalUsageReader.daily(entries: entries, localDay: LocalUsageReader.todayKey())
+        let entries = await cache.ompEntries(modifiedSince: Calendar.current.startOfDay(for: now))
+        // omp routes several models (e.g. OpenRouter) through one session log → opt into the
+        // per-model breakdown, matching Pi. Cost stays the parser's source-recorded amount.
+        return LocalUsageReader.daily(
+            entries: entries, localDay: LocalUsageReader.todayKey(), includeModels: true)
     }
 
     func fetchEnrichment() async -> ProviderEnrichment {
         let now = Date()
-        let entries = await LocalUsageCache.shared.ompEntries(
+        let entries = await cache.ompEntries(
             modifiedSince: LocalUsageReader.enrichmentScanStart(now: now))
         return .local(entries: entries, now: now)
     }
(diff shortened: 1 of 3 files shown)


chattymin:
@yourconscience Thanks for filling the gap in OMP's per-model display. We'd like to keep this PR focused on that change:

- Keep `includeModels: true`, the injectable cache, and the provider-level tests. The official OMP session writer already stores the model in `message.model`, which our current parser reads.
- Please leave out the envelope-level `model` override for now. It isn't needed for the official format, and this PR doesn't identify a fork writer that establishes the proposed precedence. This also avoids changing the parser and needing a cache-version migration in this PR.
- Please correct the “Pi is flat rate / OMP reports a real charge” explanation. Pi supports both subscription and API-key providers. OMP generally calculates cost from model pricing, with provider-reported cost overrides in some cases. Preserving recorded cost is appropriate, but it should not be described as universally billed cost.
- Resolve the conflict with current main, preserving the cost handling from #289 and the shared enrichment path, then rerun the tests on the integrated changes.

Official references:
- OMP message writer: https://github.com/can1357/oh-my-pi/blob/f97fa5c95010b62ac34c7357f9a1cae6975e12d6/packages/coding-agent/src/session/session-manager.ts#L2293-L2304
- Pi authentication options: https://github.com/earendil-works/pi/blob/71dca871bc80b6bc97be37f0ca3189399d651fff/packages/coding-agent/docs/providers.md#L1-L3
- OMP cost calculation: https://github.com/can1357/oh-my-pi/blob/f97fa5c95010b62ac34c7357f9a1cae6975e12d6/packages/catalog/src/models.ts#L138-L156

yourconscience:
Thanks for the review. All four addressed:

1. Kept `includeModels: true`, the injectable `init(cache:)` seam, and the provider-level test.

2. Dropped the envelope-level model override. `parseOmpFile` is untouched now, so attribution stays on `message.model` (what the official writer records) and there's no cache-version migration. The envelope-attribution test went with it; the provider test uses message-level models only.

3. Corrected the cost explanation. The "Pi flat-rate vs omp real charge" framing is gone. omp records its own model-price estimate in `usage.cost.total`; the parser preserves that amount and surfaces it as an estimate (`costCoverage == .estimate`), not a billed charge. The PR description is updated to match.

4. Rebased onto current main (`09fd600`), preserving #289's cost handling and the shared `.local` enrichment path. Reran on the integrated changes: build-test green, full suite 1078 tests / 11 skipped / 0 failures, coverage gate 92.94%.

One adjacent touch to flag: the `daily()` doc comment said Pi-only opt-in and now reads Pi/omp. Comment only, no logic change.

chattymin:

PR #279, opened 2026-09-07
fix: count Codex total-only turns and price Fable 5.1 cache reads

### LINKED ISSUE #277: Claude Fable 5.1 cache reads are priced 4× high — family fallback uses the Fable 5 rate
`ModelPricing.table` has no entry for `claude-fable-5-1`, so `rate(for:)` falls through to the family fallback:

```swift
if m.contains("fable")  { return .perMillion(10, 50, 12.5, 1.0) }
```

Three of those four numbers are right for Fable 5.1 — input $10, output $50, cache write $12.50. The cache **read** rate is not: Fable 5.1 lowered cache reads to **$0.25/MTok**, while the fallback applies Fable 5's $1.00/MTok. That is a 4× overcharge on the one token class that dominates agentic sessions, so displayed cost is materially overstated for anyone using `claude-fable-5-1` heavily.

### Suggested fix

```swift
"claude-fable-5-1":  .perMillion(10, 50, 12.5, 0.25),
```

### Note: `claude-opus-5` needs no change

Worth recording so this doesn't get "fixed" too — `claude-opus-5` is also absent from the table, but its `opus` fallback returns `.perMillion(5, 25, 6.25, 0.5)`, which matches Opus 5 exactly. I verified against a real day of usage (5.87M cache-read tokens): fallback and true rate both produce $6.3298, a $0.0000 difference. The version-drift fallback is doing its job there; only the Fable 5.1 cache-read rate diverges.

### LINKED ISSUE #278: Codex turns are dropped when last_token_usage components are 0 but total_tokens is set
`parseCodexToken` derives a turn's tokens from three fields of `last_token_usage`:

```swift
let inputTotal = intValue(last["input_tokens"])
let cached     = intValue(last["cached_input_tokens"])
let output     = intValue(last["output_tokens"])
let nonCachedInput = max(0, inputTotal - cached)
```

Some Codex turns emit a `token_count` event where every one of those components is `0` while `total_tokens` is set:

```json
{"timestamp":"2026-09-04T01:29:58.417Z","type":"event_msg","payload":{"type":"token_count",
 "info":{"last_token_usage":{"input_tokens":0,"cached_input_tokens":0,"cache_write_input_tokens":0,
 "output_tokens":0,"reasoning_output_tokens":0,"total_tokens":51293}}}}
```

The entry is then built with all four token classes at 0, so the turn contributes nothing.

### Impact

It is not a format-wide regression — across my 121 rollout files, 24 of 955 `token_count` events have this shape (~2.5%). But it is not evenly spread, and when a day's Codex activity happens to land entirely in those turns, the provider silently reports zero. That is what happened here: today I had exactly one Codex `token_count` event, of this shape, worth 51,293 tokens. The menu bar showed `codex: 0` with an empty `date` field, and the 51,293 tokens were dropped.

Environment: codex-cli 0.153.2, PokeTokenBar 2.5.3, macOS 26.

### Suggested fix

Fall back to `total_tokens` when the components sum to zero. This is safe because in the normal case `total_tokens` already equals exactly what the reader computes — I checked 5 normal events spanning 20K–150K tokens and `(input − cached) + cached + output == total_tokens` in every one, difference 0. `reasoning_output_tokens` is already contained in `output_tokens`, so there is no double count either way.

```swift
let nonCachedInput = max(0, inputTotal - cached)
let total = intValue(last["total_tokens"])
// components can all be 0 on some turns while total_tokens is set
let entry: Entry
if nonCachedInput + cached + output == 0, total > 0 {
    entry = Entry(..., input: total, output: 0, cacheWrite: 0, cacheRead: 0)
} else {
    entry = Entry(..., input: nonCachedInput, output: output, cacheWrite: 0, cacheRead: cached)
}
```

Attributing the fallback to `input` keeps the day's token total correct; splitting it across classes is not possible since the event carries no breakdown. If skewing the cost estimate is a concern, the alternative is to keep the token total accurate and leave cost at 0 for those turns.

This is the same class of bug as #196 (Cursor reporting zero when the local `tokenCount` is zero) — a provider whose usage silently vanishes because the reader trusts a breakdown that the source sometimes leaves empty.

Borisserz:
## Summary
- **#278:** Codex turns whose `last_token_usage` components are all `0` but `total_tokens` is set no longer vanish. We trust that total when it is the session’s only signal (no cumulative / cumulative also total-only / `last.total == cumulative.total`), or when mid-session `cumulative.total` grew since the previous event. Fork post-replay “zero-context” phantoms (`Fixtures/CodexFork/child.jsonl` L11 — filled cumulative, orphan `last.total`) stay at 0 so child totals remain `28138` / `28263`.
- **#277:** Add an exact `claude-fable-5-1` row at `$10 / $50 / $12.50 / $0.25` so the `fable` family fallback no longer prices cache reads at Fable 5’s `$1.00` (4× overstatement). `claude-opus-5` is left on the opus fallback on purpose.
- Bump `codexParserVersion` 4 → 5 so cached Codex blobs re-parse. Capture both classes in `docs/reference/defect-log.md`.

## Test plan
- [x] RED then GREEN on new guards: `testCodexTotalOnlyLastUsageCountsWhenItMatchesSessionTotal`, `…WhenCumulativeIsAbsent`, `…WhenCumulativeGrew`, plus updated Fable 5.1 asserts in `testPricingExactAndFallbackAndZero`
- [x] Phantom / fork guards still pass: `testCodexUnchangedCumulativeWithDifferentLastVectorIsPreserved` → `[110, 0]`; `testCodexManualForkFixtureKeepsOnlyPostReplayUsage` → `[0, 28138]`; sibling fork totals unchanged
- [x] Full suite: `swift test` → **908 tests, 0 failures** (11 skipped)
- [ ] Manual (optional): open a day that only has a Codex `token_count` with empty components + `total_tokens` > 0 and confirm the menu bar is non-zero; with `claude-fable-5-1` heavy cache reads, cost should be ~¼ of pre-fix

Closes #278
Closes #277

### DIFF
--- a/Sources/PokeTokenBar/Core/LocalUsageCache.swift
+++ b/Sources/PokeTokenBar/Core/LocalUsageCache.swift
@@ -78,7 +78,8 @@ actor LocalUsageCache {
     }
 
     /// fork replay 및 동일 상태 재기록 처리 변경 시 Codex blob만 재파싱한다.
-    private static let codexParserVersion = 4
+    /// v5: trust last.total_tokens when components are empty (#278).
+    private static let codexParserVersion = 5
     /// **세션 id 추출 규칙**(`session_meta` 의 id/session_id 해석·probe 종료 조건)이 바뀔 때만 올린다.
     /// resolver 변경으로 오르는 `codexParserVersion` 과 분리 — 같이 묶으면 replay 로직을 고칠 때마다
     /// 인덱스가 통째로 날아가 다음 고아 조회에서 전수 probe 가 되살아난다.
--- a/Sources/PokeTokenBar/Core/LocalUsageReader.swift
+++ b/Sources/PokeTokenBar/Core/LocalUsageReader.swift
@@ -518,6 +518,11 @@ enum LocalUsageReader {
             "\(input),\(cachedInput),\(cacheWriteInput),\(output),\(reasoningOutput),\(total)"
         }
 
+        /// Billable component sum used for Entry construction (mirrors parseCodexLine mapping).
+        var billableComponents: Int {
+            max(0, input - cachedInput) + cachedInput + output
+        }
+
         func isLower(than previous: Self) -> Bool {
             input < previous.input
                 || cachedInput < previous.cachedInput
@@ -578,6 +583,8 @@ enum LocalUsageReader {
     /// Codex 사용 엔트리. token_count 이벤트의 last_token_usage(턴 델타)를 4종 토큰으로 매핑.
     /// - input(비캐시) = input_tokens − cached_input_tokens, cacheRead = cached_input_tokens
     /// - output = output_tokens (reasoning 은 output 에 이미 포함), cacheWrite = 0
+    /// - components 가 모두 0 이고 total_tokens 만 있으면(#278): 세션이 total-only 이거나
+    ///   last.total == cumulative.total 일 때만 input 에 넣고, fork zero-context 턴은 0 유지.
     /// Codex 파일 하나를 파싱(세션 단위 — token_count 이벤트의 턴 델타). 캐시가 파일 단위로 호출.
     static func parseCodexFile(_ url: URL, fmt: DateFormatter) -> [Entry] {
         let rollout = parseCodexRollout(url, fmt: fmt)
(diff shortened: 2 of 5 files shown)


chattymin:

PR #282, opened 2026-09-10
fix: avoid UI stalls when opening large catch logs

parkjh2568:
## Summary

Opening Collection → Catch log with hundreds of entries causes a visible
stall. The eager stack creates every row, and sprite initialization
repeatedly reads cached files and constructs NSImage instances on the
main thread.

- Use LazyVStack to create catch log rows as needed.
- Share an NSImage cache across synchronous and asynchronous sprite and
  item loaders, with countLimit set to 64.
- Keep species, normal/shiny variants, PNG/GIF formats, and cache
  directories separate.
- Preserve fallback behavior without storing normal images under shiny
  keys or caching image-construction failures.
- Add rendering and cache regression tests and record the root cause
  and safeguards in the defect log.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Opening a large catch log eagerly creates every row and loads its sprites. | Rows are created lazily, reducing initial rendering work. |
| Reopening the log repeatedly reads sprite files and creates image objects. | Loaders reuse image objects while they remain cached. |

A local release-build probe with 268 entries and a 520 pt viewport
measured initial layout at approximately 783–953 ms before the change
and 22–25 ms afterward, following SwiftUI warm-up. Repeated openings
in the updated probe performed zero sprite disk reads.

These measurements cover initial layout, not end-to-end click latency.
Cold image-cache misses can still read files synchronously. Process
memory usage has not been measured; countLimit is an eviction guideline,
not a strict memory budget.

## Validation

- `swift build -Xswiftc -warnings-as-errors` passed.
- All seven new tests passed with compiler warnings treated as errors.
- `scripts/test-gate.sh` passed: 918 tests, 10 skipped, 0 failures.
  Logic-core line coverage was 91.18%.
- Inspected coverage regions for cache hits before and after await,
  image-construction failures, and fallback paths.
- Verified regression detection by temporarily restoring eager rendering
  and removing synchronous cache hits. The relevant tests failed;
  all injected defects were reverted.
- Manually verified scrolling to the oldest entries, applying and clearing
  rarity filters, switching to the Pokédex, and reopening the catch log.
- `git diff main --check` passed.
- Test fixtures use generated images and synthetic records.

## Checklist


chattymin:

PR #283, opened 2026-09-10
feat: add Aside local token usage provider

yym0329:
## Summary

Add local token-usage support for Aside through the existing provider-agnostic aggregation pipeline.

- Discover `~/.aside/u/*/state.db` and optional custom scan roots, deduplicating resolved database paths.
- Read turn-level usage metadata with read-only SQLite connections; do not read conversation bodies or credentials.
- Count input, output, cache-read, and cache-write buckets without adding `totalTokens` a second time. Fall back to the reported total when bucket values are absent.
- Sit on the shared `LocalAdditionalUsageCache` like Kiro: one scan per refresh feeds daily and enrichment, and each scan merges with previously-seen entries (`dedupKeepMax(existing + loaded)`). Aside rewrites `token_usage` in place while a turn runs and `ON DELETE CASCADE` removes a deleted session's turns, so a plain rescan would make today's total go down.
- The reader never throws. Unreadable databases (unmigrated profile, foreign `state.db` under a custom root, busy WAL recovery) are skipped; an all-failed rescan returns `[]`, which the merge treats as "nothing new" rather than zero usage. A throw from `fetchDaily` would freeze `lastUpdated` app-wide.
- Entry ids carry the file's inode so a recreated profile store (AUTOINCREMENT restarting at 1) cannot hide its new turns behind cached ids.
- Include Aside in daily, weekly, monthly, and active-block aggregation while preserving existing providers.
- Add synthetic SQLite tests (real schema: AUTOINCREMENT, `aborted_at`) and update provider-registration/layout guards.

## Type of change

- [x] New feature

## Data semantics and limitations

- Usage is attributed to the local date of the turn's last activity, `COALESCE(finished_at, last_message_timestamp)`, so a long turn that crosses midnight lands on the day its tokens were generated.
- Zero-token turns (aborted before a response) are dropped; aborted turns that already consumed tokens are counted.
- No per-model breakdown: `sessions.model` is the session's current model, not a per-turn record, so a breakdown would relabel earlier turns whenever the session switches model.
- A deleted session's already-counted usage stays counted until the scan cache is dropped (Settings save, month rollover, relaunch), matching Kiro's documented behavior.
- This adds local usage accounting, not subscription-limit tracking.

## Review history

Two independent Opus review rounds on the final shape; the second found no defects. The rounds are recorded in `docs/reference/provider-extension.md` and `docs/reference/defect-log.md` (new local sources start on the cache; reader failure semantics must be traced to `UsageStore.refresh`; test through the cache with `LocalAdditionalUsageCache(asideRootsOverride:clock:)` rather than source-text assertions).

## Validation

- `./scripts/test-gate.sh`: 925 tests, 11 skipped, 0 failures; logic-core line coverage 84.65% (gate 75%).
- Each new guard was checked by injection: turning `includeModels` back on, removing the cache merge, dropping the inode from the id, and collapsing the `finished_at` COALESCE arm each fail exactly their test.
- Read-only probe of a live Aside store: timestamps are seconds, `token_usage` buckets sum to `totalTokens`, inode is unchanged by WAL checkpoint and VACUUM. No personal database or diagnostic snapshot is included in this PR.

## Checklist

- [x] No UI source changes; the existing provider tab and aggregates display Aside

https://claude.ai/code/session_01NE6t8EQAG3aZxEZtQ7ZVfY

### DIFF
--- a/Sources/PokeTokenBar/Core/CustomScanRoots.swift
+++ b/Sources/PokeTokenBar/Core/CustomScanRoots.swift
@@ -90,6 +90,8 @@ enum CustomScanRoots {
             return LocalAntigravityUsageReader.resolvedRoots(customRootsValue: nil)
         case "opencode":
             return LocalAdditionalUsageReader.openCodeRoots(customRootsValue: nil)
+        case "aside":
+            return LocalAsideUsageReader.roots()
         case "hermes":
             return LocalAdditionalUsageReader.hermesRoots(customRootsValue: nil)
         case "cursor":
--- a/Sources/PokeTokenBar/Core/CustomScanRoots.swift
+++ b/Sources/PokeTokenBar/Core/CustomScanRoots.swift
@@ -91,7 +91,7 @@ enum CustomScanRoots {
         case "opencode":
             return LocalAdditionalUsageReader.openCodeRoots(customRootsValue: nil)
         case "aside":
-            return LocalAsideUsageReader.roots()
+            return LocalAsideUsageReader.roots(customRootsValue: nil)
         case "hermes":
             return LocalAdditionalUsageReader.hermesRoots(customRootsValue: nil)
         case "cursor":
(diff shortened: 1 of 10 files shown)


chattymin:

PR #291, opened 2026-09-11
feat: tell shoppers what a released Pokémon keeps

LuceteYang:
## Why

I bought an egg expecting to lose the Pokémon I was raising — and specifically expecting it to disappear from my Pokédex. Reading the code afterwards, I was wrong on both counts that matter.

The card says only *"Send off your current Pokémon and start fresh with a new egg."* Nothing follows it, so "send off" reads as "lose it", and the shiny warning sitting right next to it reinforces that something irreversible is happening.

What actually happens:

- `buyEgg()` appends `releasedDexEntry(from:)` before clearing the active Pokémon, so the species stays in the Pokédex. `FreshEggTests.testReleasedSpeciesStaysInTheDex` locks this in.
- `chooseBase()` halves the hatch weight of any base whose line you have already completed. That check reads `collectedFinals`, which only `graduate()` fills — releasing leaves it untouched, so a released Pokémon keeps its full weight and can hatch again at the same odds.

The one thing you do give up is the growth progress. The shop says none of this, so the card reads scarier than the feature is.

Egg card, before and after

And in Korean, where the description fits on one line so the note costs two:

Egg card in Korean, before and after

<sub>Rendered the same way as `scripts/gen-settings-screenshots.py` does its assets — the popover cannot be opened from a script, so the card is drawn in HTML with the colours and metrics sampled from `assets/screenshot-shop.png` / `-ko.png`.</sub>

## What changes

Adds `eggReleaseNote` right under `eggDescription`:

> A released Pokémon stays in your Pokédex and can hatch again at the same odds — only the growth progress is lost.

It goes in the description column rather than the price row, which belongs to the price, the button and `eggShopLockedHint`. Rendered only while there is a Pokémon to release. Translated for all seven languages.

Re-incubation is deliberately left out — `eggDescription` already says the new egg starts from scratch, and repeating it pushed the German string to three lines.

## Testing

`swift build` and `swift test` (1048 tests) pass.

### DIFF
--- a/Sources/PokeTokenBar/Core/Localization.swift
+++ b/Sources/PokeTokenBar/Core/Localization.swift
@@ -991,6 +991,24 @@ struct L {
           "Disponível quando seu ovo atual chocar.",
           "Verfügbar, sobald dein aktuelles Ei geschlüpft ist.")
     }
+    /// 놓아주면 무엇이 남는지 — `eggDescription` 바로 아래 줄(EggCard).
+    ///
+    /// 경고가 아니라 **안심**이 목적이다. `eggDescription` 이 "놓아준다"까지만 말해서 도감까지 잃는다고
+    /// 읽히지만, `buyEgg` 는 `releasedDexEntry` 로 도감에 남긴다.
+    ///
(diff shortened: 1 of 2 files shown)


chattymin:

PR #298, opened 2026-09-13
feat: show a pace marker on quota bars

LuceteYang:
## Summary

Quota bars only warned once a window was nearly exhausted. They showed how much
was spent, but never whether that spending was fast for how far into the window
it was — the difference between "62% with four hours left" and "62% with twenty
minutes left".

Draw a tick at the point an even burn would have reached by now. Fill past the
tick means the window will run out before it resets.

The tick is sized against the *painted track*, not the bar's layout box:
`ProgressView` reserves 12pt of height while painting a 6pt track, so framing the
tick against the overlay's geometry made it 16pt — taller than the bar it marks,
reading as a staple laid over the bar rather than a mark on it. It now spans the
6pt track plus 2pt of overhang per side, 2.5pt wide with rounded ends. 2.5pt
lands on exactly 5px at 2x, so the tick stays crisp instead of smearing across a
half pixel the way a 1pt hairline can.

Rebased onto v2.5.4. #292 had since folded every quota row into a shared
`quotaRow`, so the marker now hangs off that one helper instead of four
near-duplicate rows — `span` is an optional parameter there, defaulting to nil.

The marker needs the window length, which every provider reports its own way, so
`LimitWindowSpan` collects those mappings (Codex: explicit `windowDurationMins`,
Antigravity: window name, Claude: entry `kind` or legacy field name) and each row
passes a plain `TimeInterval` down. Rows whose length is unknown — the Codex
personal spend limit — simply get no marker, and no provider literal enters the
shared path.

Two deliberate restraints:

- **No timer.** A 5-hour window moves 0.33%p per minute, so the existing refresh
  interval and reopening the popover already carry the tick along. The always-on
  animation budget is untouched.
- **An out-of-range elapsed fraction hides the marker instead of clamping.** A
  5-hour window starts at the first request, so an idle window can carry a stale
  `resets_at`; pinning that to either end would claim the window just started or
  just ended. A tick in the wrong place is worse than no tick.

Marker position runs through the same `displayPercent` conversion as the fill, so
"remaining" display mode cannot flip one without the other.

Hovering a row explains the tick: *"Pace — an even burn across this window would
sit at 48% now."* (7 languages.) The copy deliberately avoids calling this an
"optimal" or "recommended" amount — it is a risk signal, not advice to use less,
and there is no reason to spend a whole window on a quiet day.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
|  |  |

Both frames are captures of the running app, taken minutes apart on the same
machine and cropped identically. Before, the rows read the same in kind — just
percentages and a fill. After, every row also says where an even burn would have
reached by now: the 5-hour session is at 46% with the tick far ahead of the fill
(plenty of headroom), while the weekly rows sit much closer to their ticks.

## Validation

- Full gate: 1,086 tests, 13 skipped, 0 failures. Core line coverage 92.88%
  (75% minimum). The extra skip is the screenshot generator, which only runs
  with `PTB_SCREENSHOT_DIR` set.
- New pure tests cover every window-length mapping (Codex minutes incl. 0 and
  negative, Antigravity `window` plus `bucketId` fallback, Claude `kind`), the
  elapsed fraction at window start/middle/end, and both display modes.
- The rendering test reads painted pixels rather than the view tree: a SwiftUI
  shape never becomes an `NSView`, so a pure-function test alone would not catch
  the view passing a hardcoded display mode.
- A second rendering test measures the tick's painted *height* against the
  painted track. The position test only looks for the darkest column, so it
  passes whichever box the height came from — which is exactly how the oversized
  tick got through. Putting the old frame back fails it (15pt against 10pt).
- Confirmed the new guards fail when the defect is injected, rather than only
  passing: hardcoding `.used` in the marker (rendering test caught 50 vs 150),
  clamping the elapsed fraction instead of hiding (stale-`resets_at` assertions
  caught it), and dropping the line width from the offset (the tick fell outside
  the bar at window end and stopped rendering entirely).
- Inspected the new branches with `llvm-cov show --show-regions`; no unexecuted
  regions. Line coverage alone was not treated as evidence.

The tick uses `.primary` at 55% so it stays legible over both the filled and
empty parts of the bar, in light and dark, and its size is expressed in points
relative to the track rather than to any one provider's bar. The fill itself is
tinted by utilisation — green, orange past the warning threshold, red past the
critical one — so the tick always has a coloured bar to sit against.

(Worth knowing for anything that generates screenshots from tests: an offscreen
test process never becomes active, so AppKit draws the progress indicator
unemphasized and the fill comes out grey. The images above are captures of the
running app for that reason.)

## Checklist

### DIFF
--- a/Sources/PokeTokenBar/Core/Localization.swift
+++ b/Sources/PokeTokenBar/Core/Localization.swift
@@ -52,6 +52,17 @@ struct L {
     var weeklySonnet: String { t("주간 Sonnet", "Weekly Sonnet", "週間 Sonnet", "Sonnet semanal", "Sonnet hebdo", "Sonnet semanal", "Sonnet – wöchentlich") }
     var claudeCurrentBlock: String { t("Claude 현재 5h 블록", "Claude current 5h block", "Claude 現在の5hブロック", "Bloque actual de 5h de Claude", "Bloc 5 h actuel de Claude", "Bloco atual de 5h do Claude", "Aktueller 5-Stunden-Block von Claude") }
     var reset: String { t("리셋", "Reset", "リセット", "Reinicio", "Réinit.", "Renovação", "Zurücksetzen") }
+    /// 페이스 눈금 툴팁. "적정 사용량"으로 부르지 않는다 — 덜 쓰라는 권고가 아니라 "이 속도면 리셋
+    /// 전에 소진된다"는 기준선이고, 창을 다 채워 쓸 이유가 없는 날에 규범으로 읽히면 안 된다.
+    func paceHint(_ percent: String) -> String {
+        t("페이스 — 창 시간만큼 균등하게 썼다면 지금 \(percent)입니다.",
+          "Pace — an even burn across this window would sit at \(percent) now.",
+          "ペース — この枠を均等に使っていれば今は \(percent) です。",
+          "Ritmo: un consumo uniforme en esta ventana estaría en \(percent) ahora.",
+          "Rythme — une consommation régulière sur cette fenêtre serait à \(percent).",
+          "Ritmo — um consumo uniforme nesta janela estaria em \(percent) agora.",
+          "Tempo – bei gleichmäßigem Verbrauch in diesem Fenster wären es jetzt \(percent).")
+    }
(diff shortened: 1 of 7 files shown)


chattymin:

PR #302, opened 2026-09-13
fix: re-read Keychain on limit refresh and toggle the floating pet from the footer

### LINKED ISSUE #300: Refresh limit token cache doesn't re-read Keychain when the previous token is still valid (stale account after switching)
## Environment
- PokeTokenBar 2.5.4
- macOS 26.6 (25G72), Apple Silicon
- Keychain-only install (no `~/.claude/.credentials.json`)
- Claude Code, Team plan, switching between two email accounts on the same Team

## Problem

After switching Claude Code accounts via `/logout` → `/login` (same Team, two emails), the official
5h/weekly limits in PokeTokenBar keep showing the **previous** account indefinitely, even after
repeatedly pressing Settings → **Refresh limit token cache**.

- Popover keeps showing `Account: edoardo@memoraiz` after switching to `hello@memoraiz`.
- `fiveHour`/`sevenDay` stay frozen for many refresh cycles while the local token counter (feeding
  companion EXP) keeps climbing normally — the two data paths clearly disagree.

## Root cause (traced in OAuthLimitsProvider.swift)

`UsageStore.refreshLimitTokenFromKeychain()` calls `limitsProvider.fetch(allowKeychainPrompt: true)`,
which calls `OAuthAccessTokenCache.accessToken(allowKeychainPrompt: true)` — with the default
`bypassCache: false`:

```swift
if !bypassCache, let cachedCredential, !cachedCredential.isExpired {
    return cachedCredential.accessToken
}
```

`cachedCredential` is only cleared by `invalidate()`, called only from the 401/403 retry path. A
second, independently-valid Team account's token never gets a 401/403 from `api.anthropic.com`, so
it's returned from memory forever — Keychain is never queried again after the first successful read,
no matter how many times the refresh button is pressed.

This also makes the log misleading: `UsageStore.swift` logs `"limits refreshed from keychain by user
action"` unconditionally on success, even when the token actually came from the in-memory cache and
Keychain was never touched. My own log shows exactly this pattern for hours after switching accounts.

Different bug from #227/#228 (fixed the `.credentials.json` file path) and #232 (fixed picking the
right entry among *multiple* Keychain items) — this install has no credentials file, and presumably
one relevant Keychain item, so neither prior fix applies here.

## Steps to reproduce

1. Keychain-only install. Sign into Claude Code as Account A (Team). Press **Refresh limit token
   cache** once so the popover shows Account A.
2. `/logout`, then `/login` as Account B (different email, same Team, so Account A's token stays
   valid server-side).
3. Press **Refresh limit token cache** again (any number of times).

**Expected:** limits and account label switch to Account B.
**Actual:** limits and account label stay on Account A indefinitely.

## Suggested fix

Have the explicit user-action path force `bypassCache: true` — it's already an intentional user
action, same category as the existing 401/403-triggered bypass retry, so there's no
privacy/prompt-avoidance reason to prefer the cache here.

### LINKED ISSUE #301: Floating pet: quick on/off toggle outside Settings (Hide is quick, Show isn't)
## Problem

Turning the floating pet **off** is quick — right-click it → Hide (`FloatingPetPanel.swift`,
`floatingPetMenuHide` → `store.floatingPetEnabled = false`). Turning it back **on** requires opening
Settings, because once it's hidden there's nothing left to right-click.

The popover footer already has a small icon row (refresh / gear / power) — a floating pet toggle
there would make both directions equally quick.

## Suggestion

A small toggle icon in the popover footer, bound directly to `floatingPetEnabled` — the same
variable the Settings checkbox and the pet's own "Hide" menu item already use, so no new state to
introduce.

Borisserz:
## Summary

- **#300:** Settings / popover **Refresh limit token cache** used to return the in-memory Claude token whenever it was still unexpired. After `/logout` → `/login` to another email on the same Team, that token stays valid, so Keychain was never opened and the official 5h/weekly bars stayed on the previous account. The user-action path now passes `bypassCache: true`. Auto-poll still does not touch Keychain. The same one-line rule is applied to Antigravity's user refresh, which had the same still-valid-cache hole.
- **#301:** Hide was a right-click on the pet. Show only lived in Settings, because a hidden pet has nothing to click. The popover footer now has an eye toggle bound to the existing `floatingPetEnabled` flag, on every tab, with the existing Show / Hide strings.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Floating pet can be hidden from its menu, but turning it back on requires Settings. | Footer (next to Settings / Quit, on every tab) has an eye icon. Shown: “Hide floating pet”. Hidden: “Show floating pet”. Same flag as the Settings checkbox. |
| Refresh limit token cache can leave official limits on the previous Claude account when that token has not expired. | The same button re-reads Keychain and the account label / 5h / weekly bars follow the account now stored there. |

## Checklist

Closes #300
Closes #301


chattymin:

PR #304, opened 2026-09-13
fix: price the Claude 5 family so Claude usage costs are no longer "Unavailable"

### LINKED ISSUE #303: Claude usage cost shows "Unavailable": Claude 5 family missing from the pricing table
## Symptom

Claude usage shows **"Unavailable" / `$—`** instead of a cost, for every current Claude Code user. The token counts are correct; only the cost is missing.

## Root cause

`claude-opus-5` and `claude-sonnet-5` are absent from `ModelPricing.table` (`Sources/PokeTokenBar/Core/ModelPricing.swift`).

Chain:

1. Claude Code `*.jsonl` records carry no explicit cost field, so `Entry.explicitCost` is `nil`.
2. `ModelPricing.estimatedCost(model: "claude-opus-5", …)` returns `nil` — no table row, and the family fallback no longer exists.
3. `LocalUsageReader.Bucket.add` takes the last branch and merges `.unavailable` into `costCoverage`.
4. `UsageCost.text` hits `coverage.unknown && !coverage.hasKnown` and renders `l.costUnavailable`.

Because the Claude 5 family is the current generation, effectively *all* Claude usage lands on that branch — the coverage is unknown-only, not partial, so no amount is shown at all.

Measured on a real profile: 12,492 of 13,318 `"model"` occurrences over the last 7 days were `claude-opus-5`; for the current day it was 417 of 417.

## Regression origin

Before #289 (`fix: distinguish estimated and unavailable usage costs`, shipped in v2.5.4), `ModelPricing.rate(for:)` had a model-family fallback:

```swift
if m.contains("opus")   { return .perMillion(5, 25, 6.25, 0.5) }
```

`claude-opus-5` was priced solely through that fallback. #289 removed the fallback — correct in itself, since #277 showed the fallback mispriced Claude Fable 5.1 by 4× — but no exact rows were added for the Claude 5 family to replace it.

This was foreseen and written down. `docs/reference/defect-log.md:217`:

> 폴백이 우연히 맞는 모델(예: `claude-opus-5` → opus 폴백)은 건드리지 않는다

The guidance was to add exact rows whenever the fallback stops covering a model. The fallback was removed without that step.

## Why no test caught it

- `ModelPricingTests.testCurrentOpenAIStandardRates` asserts GPT rows only. Nothing asserts that the Claude models actually read from logs are priced.
- `ModelPricingTests.testUnknownNamesNeverBorrowFamilyPrices` encodes "unknown model → `nil`" as *correct* behaviour. `claude-opus-5` silently joined that class, so the suite confirms the regression instead of reporting it.
- Every other test uses models that are in the table (`claude-fable-5-1`, `claude-opus-4-8`, `claude-sonnet-4-6`, `gemini-2.5-pro`), so all of them pass.
- Line coverage does not help: the `.unavailable` branch *is* executed by existing tests. This is the "coverage number is not evidence" case from `CLAUDE.md`.

## Class sweep

Models observed across local provider logs that currently resolve to `nil`:

| Model | Priced? | Note |
| --- | --- | --- |
| `claude-opus-5` | no | current default model |
| `claude-sonnet-5` | no | present in logs |
| `gemini-3-pro` | no | separate question, not covered here |
| `codex-auto-review` | no | not a real model id, correct as-is |

The whole Claude 5 family is missing; the remaining Claude rows (`claude-opus-4-8`, `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5-20251001`, `claude-fable-5`, `claude-fable-5-1`) were verified against the official pricing page and are correct.

## Fix

Add exact rows, taken from https://platform.claude.com/docs/en/about-claude/pricing (base input / 5m cache write / cache read / output):

| Model | Input | Output | 5m cache write | Cache read |
| --- | --- | --- | --- | --- |
| Claude Opus 5 | $5 / MTok | $25 / MTok | $6.25 / MTok | $0.50 / MTok |
| Claude Sonnet 5 | $2 / MTok | $10 / MTok | $2.50 / MTok | $0.20 / MTok |

```swift
"claude-opus-5":   .perMillion(5, 25, 6.25, 0.5),
"claude-sonnet-5": .perMillion(2, 10, 2.5, 0.2),
```

## Permanent capture

A row-by-row table test is not enough — it would have been written against the same stale list. Add a regression test that asserts the *currently shipping* model identities resolve to a non-`nil` estimate, i.e. the inverse of `testUnknownNamesNeverBorrowFamilyPrices`. That test fails the moment a new generation ships unpriced, which is exactly the trigger condition here, and it is the guard the removal in #289 lacked.

Also record the class in `docs/reference/defect-log.md`: removing a fallback is a behaviour change for every input the fallback used to serve, and those inputs must be enumerated before the fallback goes.

florianbehrend:
Fixes #303.

## Problem

Claude usage rendered **"Unavailable" / `$—`** instead of a cost. Token counts were correct; only the amount was missing.

`claude-opus-5` and `claude-sonnet-5` were absent from `ModelPricing.table`, so:

1. Claude Code `*.jsonl` carries no explicit cost field → `explicitCost == nil`.
2. `ModelPricing.estimatedCost` returns `nil` — no row, and no family fallback any more.
3. `LocalUsageReader.Bucket.add` falls through to `costCoverage.merge(.unavailable)`.
4. `UsageCost.text` hits `coverage.unknown && !coverage.hasKnown` → `l.costUnavailable`.

Because the Claude 5 family is the current generation, the coverage is unknown-*only* rather than partial, so no amount is shown at all. Measured on a real profile: 12,492 of 13,318 `"model"` occurrences in the last 7 days were `claude-opus-5`; 417 of 417 for the current day.

## Regression origin

Before #289 (shipped in v2.5.4), `rate(for:)` had a model-family fallback:

```swift
if m.contains("opus") { return .perMillion(5, 25, 6.25, 0.5) }
```

`claude-opus-5` was priced solely through it. Removing the fallback was correct — #277 showed it mispriced Fable 5.1 by 4× — but no exact rows replaced it for the models it had been covering. `docs/reference/defect-log.md:217` had explicitly flagged `claude-opus-5` as depending on that fallback.

## Why the suite stayed green

`testPricingExactAndFallbackAndZero` contained:

```swift
// An unverified future model remains unpriced.
XCTAssertEqual(ModelPricing.rate(for: "claude-opus-5"), .zero)
```

The suite did not merely miss the defect — it asserted it as intended behaviour. `testUnknownNamesNeverBorrowFamilyPrices` reinforces the same direction, and every other pricing test uses models that are in the table. Line coverage is no help either: the `.unavailable` branch *is* exercised by existing tests, so the gate passes.

## Changes

**`ModelPricing.table`** — exact rows, from https://platform.claude.com/docs/en/about-claude/pricing (base input / output / 5m cache write / cache read):

| Model | Input | Output | 5m cache write | Cache read |
| --- | --- | --- | --- | --- |
| Claude Opus 5 | $5 / MTok | $25 / MTok | $6.25 / MTok | $0.50 / MTok |
| Claude Sonnet 5 | $2 / MTok | $10 / MTok | $2.50 / MTok | $0.20 / MTok |

I verified the remaining Claude rows against the same page while I was there; they are all correct.

**`testPricingExactAndFallbackAndZero`** — now asserts the real rates for both models. The "remains unpriced" case moves to `claude-opus-6`, an identity that does not exist, so the assertion can no longer capture a shipping model.

**`testModelsReadFromProviderLogsAreAllPriced`** (new) — the inverse of `testUnknownNamesNeverBorrowFamilyPrices`. That test guards against a *wrong* price; this one guards against *no* price, over the identities actually read from provider logs, using a realistic bucket split (cache creation and cache reads non-zero, as every Claude Code request logs). A table test listing only the rows already present could not have caught this.

**`docs/reference/defect-log.md`** — records two classes: removing a fallback is a behaviour change for every input it used to serve (enumerate them from real logs before deleting), and an "unknown → nil" assertion naming a current-generation model locks the defect in.

## Verification

- `swift build` passes.
- The new guard was checked against an injected defect, not just for a green run: compiled against the pre-fix `ModelPricing.swift` from `main` it reports `FAIL: claude-opus-5, claude-sonnet-5, rate:claude-opus-5, rate:claude-sonnet-5`, and `PASS` against the fixed table.
- `swift test` could not be run locally — this machine has only CommandLineTools, no `Xcode.app`, so `XCTest` is unavailable (`error: no such module 'XCTest'`). The suite runs in CI on this PR.
- Checked that the new rows do not disturb other suites: Copilot fixtures use `claude-opus-5` but set `costUnavailable: true`, which short-circuits the estimate branch, and Cursor uses the distinct key `claude-opus-5-thinking-high`.


florianbehrend:
Flagging an overlap: **#313 edits the same two table rows.** It was opened a day after this one and nobody has put the two side by side yet, so right now there are two open PRs touching the same lines.

The core is identical — same two ids, same rates ($5/$25/$6.25/$0.50 and $2/$10/$2.50/$0.20), and both correct the `XCTAssertEqual(ModelPricing.rate(for: "claude-opus-5"), .zero)` assertion in `LocalUsageReaderTests` that had pinned the defect as intended behaviour. What differs is what ships alongside:

| | #304 (this) | #313 |
| --- | --- | --- |
| Table rows + `.zero` fix | yes | yes |
| Regression guard | `testModelsReadFromProviderLogsAreAllPriced` — asserts every identity actually read from provider logs resolves to a price (8 ids) | `testCurrentAnthropicClaude5Rates` — the two new models |
| `docs/reference/defect-log.md` | two classes recorded | — |

I'd suggest landing this one and closing #313 as a duplicate, for two reasons that aren't about authorship:

**The guard has to catch the *next* occurrence, not this one.** A test naming `claude-opus-5` and `claude-sonnet-5` passes forever once those rows exist — it cannot fail when the following generation ships unpriced, which is precisely how this defect reached a release. The guard here enumerates the identities actually read from provider logs and fails the moment any of them has no row. I verified it against an injected defect rather than only for a green run: compiled against the pre-fix `ModelPricing.swift` from `main` it reports `FAIL: claude-opus-5, claude-sonnet-5, rate:claude-opus-5, rate:claude-sonnet-5`, and `PASS` after.

**The regression has a documented cause worth recording.** These two models were priced *only* through the `contains("opus")` family fallback that #289 removed; `docs/reference/defect-log.md:217` had explicitly named `claude-opus-5` as depending on it. The defect-log entries in this PR capture that as a class — removing a fallback is a behaviour change for every input it served, and an "unknown → nil" assertion naming a current-generation model locks the defect in. Per the repo's defect protocol that capture is part of the fix, not an extra.

chattymin:

PR #307, opened 2026-09-14
fix: clamp extra-provider token counts so a corrupt log cannot trap

### LINKED ISSUE #306: Extra-provider parsers trap on oversized token counts and crash the refresh
## Symptom

A single corrupt or hand-edited usage log kills the app on every refresh. The process dies with `SIGTRAP` (`Double value cannot be converted to Int64` / Swift integer overflow) until that file is removed. Save recovery does not help, because decoding succeeded.

Reproduced with a `1e30` token count in:

- Cursor chat bubbles and the Cursor dashboard usage API
- OpenCode message JSON
- Hermes `sessions` rows (`output_tokens + reasoning_tokens`)
- Kiro `request_start_timestamp_ms`

## Root cause

`LocalUsageReader` already clamps external numbers to `maxParsedTokenValue` so a later add cannot overflow. The extra-provider extractors were not part of that sweep:

- `LocalAdditionalUsageProvider.intValue` and `CursorUsageAPI.intValue` use `NSNumber.intValue`, which saturates at `Int.max`. The next `input + output` then traps.
- `columnInt` passes a SQLite `INTEGER` of `2^63-1` through, and Hermes adds two of those columns.
- Kiro calls `Int64(Double)` on a timestamp. A finite value past `Int64.max` traps.

## Expected

A bad number is clamped. The refresh continues. Normal counts are unchanged.

Borisserz:
## Summary

Fixes #306.

- Extra-provider parsers (`Cursor`, OpenCode, Hermes, Kiro, Cursor dashboard API) saturated oversized log numbers at `Int.max`, then trapped on the following add. The same file killed every refresh.
- Those extractors now use the same `maxParsedTokenValue` ceiling as `LocalUsageReader`. Kiro timestamps no longer call `Int64(Double)` past `Int64.max`.
- Normal counts are unchanged. Regression tests are split per provider so one trap cannot hide the next case.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## Test plan

- [x] `testCursorHugeTokenCountsDoNotTrapOnTheAdd`
- [x] `testOpenCodeHugeTokenCountsDoNotTrapOnTheBucketSum`
- [x] `testHermesHugeSessionTokensDoNotTrapOnTheReasoningAdd`
- [x] `testHugeRequestTimestampDoesNotTrap`
- [x] `testParseCursorUsageEventClampsHugeTokenCounts`
- [x] Existing token-mapping tests for Hermes, Kiro, and Cursor still pass
- [ ] CI `swift test`


chattymin:

PR #309, opened 2026-09-14
fix: price gemini-2.5-flash-lite so its cost is not Unavailable

### LINKED ISSUE #308: Gemini 2.5 Flash-Lite usage cost shows Unavailable
## Symptom

Gemini CLI sessions that use `gemini-2.5-flash-lite` show token counts, but the cost is **Unavailable**. Flash and Pro on the same machine still get a dollar estimate.

## Root cause

`ModelPricing.table` has exact rows for `gemini-2.5-flash` and `gemini-2.5-pro`, and unknown model names intentionally return no estimate (so a cheaper SKU is not billed as Flash). `gemini-2.5-flash-lite` was never added, so `estimatedCost` is `nil`.

Official standard text rates (storage and audio excluded, same as the other Gemini rows): $0.10 input / $0.40 output / $0.01 cache read per million tokens.

https://ai.google.dev/gemini-api/docs/pricing

## Expected

An exact `gemini-2.5-flash-lite` row, including the `models/` prefix the CLI sometimes writes. Other lite ids that are not in the table stay Unavailable.

Borisserz:
## Summary

Fixes #308.

- `gemini-2.5-flash-lite` was missing from the price table, so Gemini CLI sessions on that model showed a cost of Unavailable.
- Adds the official standard text rates only: $0.10 input / $0.40 output / $0.01 cache read per million tokens. Audio, cache-write, and storage stay unavailable, same as the other Gemini rows.
- Does not add a family fallback. `gemini-3-flash-lite` and other unknown lite ids still return no estimate.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## Test plan

- [x] `ModelPricingTests/testGemini25FlashLiteUsesOfficialTextRates` failed on missing row, then passed
- [x] `ModelPricingTests` and `GeminiUsageTests/testGeminiPricing` pass
- [ ] CI `swift test`


chattymin:

PR #311, opened 2026-09-14
fix: show a skipped release in Settings instead of "up to date"

### LINKED ISSUE #310: "Later" on the update banner skips the version permanently, and Settings then reports the app is up to date
**Environment**
- PokeTokenBar 2.5.3 (Homebrew cask), macOS 26.5 (25F71), Apple Silicon
- AI CLI in use: Claude Code

**Steps to reproduce**
1. On 2.5.3, wait for the update banner for 2.5.4 to show up in the popover.
2. Click **Later**.
3. Open Settings → Updates → **Check now**.

**Expected:** the app offers 2.5.4 again — immediately, or on some later check.

**Actual:** the banner never comes back, and Settings reports *"You're on the latest (v2.5.3)"* while 2.5.4 is released.

**What I found in the source**

`updateLater` ("Later", `Localization.swift:882`) is wired to `skipCurrent()` (`PopoverView.swift:100`), which writes the version into `skippedUpdateVersion` (`UpdateChecker.swift:51`). Repo-wide that key has exactly two references — one read, one write — so nothing ever clears it again. `checkForUpdate()` then keeps `available` at `nil` because `latest == skipped` (`UpdateChecker.swift:43`), and the Settings row falls through to its `else` branch and prints `upToDate` (`SettingsView.swift:364`).

The skip itself works as designed. What caught me out is that the up-to-date message doesn't distinguish *"no newer release"* from *"newer release, skipped earlier"* — so the check quietly reports something that isn't the case.

Workaround for anyone who runs into it, with the app quit (it writes its preferences back on exit):

```
defaults delete io.github.chattymin.poketokenbar skippedUpdateVersion
```

**Possible directions**
- *Let "Later" mean later:* don't set `skippedUpdateVersion`, and offer the version again on the next check — or store a "remind after" timestamp instead.
- *Keep the skip, rename the button:* "Skip this version", and have Settings still surface that a newer version exists, so the state stays discoverable.

Either way, it would help if the "You're on the latest" line didn't show up for a version that was skipped.

Borisserz:
## Summary

Fixes #310.

- The banner button labeled **Later** called `skipCurrent()`, which permanently hid that version. Settings treated `available == nil` as "You're on the latest".
- Keep the skip. Rename the button to **Skip this version**, and show the skipped release in Settings with **Update** and **Show again**.
- A newer tag than the skipped one still opens the banner. **Show again** restores the banner without `defaults delete`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## UI changes

| Before | After |
| ------ | ----- |
| Banner: **Later** sits beside **Update** in one row. Clicking it hides that version forever. | Banner: **Skip this version** on its own row, so the longer label fits. It still hides the banner for that version only. |
| Settings → Check now: skipped release prints "You're on the latest". No way back except `defaults delete`. | Settings → Check now: "You skipped vX", plus **Show again** and **Update**. A later release still uses the update-found row. |

## Checklist

- [x] `swift build` and `swift test` pass locally (1079 tests, 13 skipped, 0 failures)

## Test plan

- [x] `UpdateCheckerTests/testSkippedReleaseStaysVisibleAndANewerOneReturnsToTheBanner` fails if a skipped release is dropped, then passes
- [x] `UpdateCheckerTests/testShowAgainRestoresTheBannerAndUpdateUsesTheSkippedRelease` passes
- [x] `LocalizationInterpolationTests` keeps the skipped-version placeholder in all 7 languages
- [x] `swift build` and `swift test` locally: 1079 passed, 13 skipped, 0 failed
- [ ] CI `swift test`


chattymin:

PR #313, opened 2026-09-14
fix: price Claude Opus 5 and Sonnet 5 in the model rate table

Oreo-Mcflurry:
## Summary

`ModelPricing.table` had no row for Claude Opus 5 (`claude-opus-5`) or Claude Sonnet 5 (`claude-sonnet-5`), and neither id resolves through `aliases`. `ModelPricing.estimatedCost()` therefore returned `nil` for both — the "cannot be priced / unavailable" signal — so usage recorded under those models contributed nothing to the today / week / month cost totals. For a log dominated by either model, today's cost read far lower than reality, or close to $0.

Fixes #312.

### Rates

Source: https://claude.com/pricing, checked 2026-09-14.

| Model | Input $/MTok | Output $/MTok | Cache write $/MTok | Cache read $/MTok |
| --- | --- | --- | --- | --- |
| `claude-opus-5` | 5 | 25 | 6.25 | 0.5 |
| `claude-sonnet-5` | 2 | 10 | 2.5 | 0.2 |

Cache write and cache read follow the same convention as the neighbouring `claude-opus-4-8` / `claude-sonnet-4-6` rows: `cacheWrite = input × 1.25`, `cacheRead = input × 0.1`. These are exact-match rows rather than a family fallback, per the #277 note in `docs/reference/defect-log.md` ("패치·마이너가 단가 한 칸만 바꿀 수 있으면 정확 매칭 행을 표에 추가").

## Scope

Deliberately minimal — no refactoring, no formatting changes, no changes to `estimatedCost()` or any other logic.

1. **`Sources/PokeTokenBar/Core/ModelPricing.swift`** — two table rows plus a source/date comment in the style of the existing "checked YYYY-MM-DD" comments.
2. **`Tests/PokeTokenBarTests/ModelPricingTests.swift`** — one new test, `testCurrentAnthropicClaude5Rates`, asserting both rate rows and that `estimatedCost()` now returns a value instead of `nil`. Verified to fail before the table change:
   ```
   ModelPricingTests.swift:16: XCTAssertEqual failed:
     ("ModelRate(input: 0.0, output: 0.0, cacheWrite: 0.0, cacheRead: 0.0)") is not equal to
     ("ModelRate(input: 5e-06, output: 2.5e-05, cacheWrite: 6.25e-06, cacheRead: 5e-07)")
   ModelPricingTests.swift:19: XCTUnwrap failed: expected non-nil value of type "Double"
   ```
[...]

PR #318, opened 2026-09-16
fix(claude): prefer reported cost-state over price-table estimates

DenisSeller:
## Summary

Claude Code writes its own cost ledger into each session file — cumulative
`type:"cost-state"` records carrying per-model `costUSD` inside `modelUsage`.
`LocalUsageReader` never read it, so Claude usage was always priced from
`ModelPricing`, and any model without a table row aggregated as unavailable
even though an exact amount was already present in the same file.

This parses the last cost-state record per file and spreads each model's
session total across that model's entries in proportion to tokens, feeding the
existing `Entry.explicitCost` that `Bucket.add` already prefers over an
estimate. Where no ledger exists, behaviour is unchanged: table estimate, else
unavailable.

### Relationship to #304 and #313

Those PRs add `claude-opus-5` / `claude-sonnet-5` rows to the table and should
land regardless — this change does not replace them. It sits *ahead* of the
table, so the two compose: a reported amount wins, and the table stays the
fallback for sessions with no ledger.

They are also measurably correct for standard context. Checking both PRs'
proposed rates against the reported ledger on a local 128-session sample:

| model | vs. reported ledger |
| --- | --- |
| `claude-sonnet-5` | +0.3% |
| `claude-opus-5[1m]` | **−6.4%** |

The gap is the context-window variant. `modelUsage` keys 1M-context traffic as
`claude-opus-5[1m]`, which prices differently from `claude-opus-5`, and a
single table key cannot express both. Reading the reported amount closes that
without a new row per model or per context tier.

Worth flagging for #303, which states that Claude `*.jsonl` records carry no
explicit cost field: they do — `cost-state` has been there, with
`hasUnknownModelCost` false across all 128 sessions sampled.

### Attribution rules

- The ledger is cumulative, so only the final record describes the session.
- `modelUsage` keys carry a context suffix that `message.model` never has, so
  matching is on the base id and both variants pool onto it.
- The session sum stays exact — only the split *within* a session is inferred,
  which is why the amount keeps `.source` coverage rather than `.estimate`.
- Cost for a model that produced no parsed entry is **dropped, not reassigned**,
  so an unattributable amount can never inflate another model's day. On the
  sample above that is a deliberate 0.7% undercount.
- `claudeParserVersion` is added because Claude blobs previously invalidated on
  mtime/size alone; without it, entries cached under the old rule would keep
  their cost-free values until each file happened to change.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## Checklist

`swift test`: 1086 executed, 0 failures. New `ClaudeReportedCostTests` covers
context-suffix matching, cumulative last-record-wins, exact day-split across a
session boundary, unattributable-model cost, zero-token records, reported-over-
estimate precedence, and the unchanged no-ledger fallback. No files under
`Sources/PokeTokenBar/UI/` are touched.

https://claude.ai/code/session_01KwKz8zsXemuCwftTX2gN3k


chattymin:

PR #321, opened 2026-09-16
fix: stamp update-check cooldown only after a successful fetch

### LINKED ISSUE #320: Failed update check starts a 30-minute cooldown and blocks retries
## Summary

`UpdateChecker.check()` stamps `lastChecked` **before** the GitHub request finishes. If that request fails (offline, timeout, non-200, bad JSON), the next automatic check — including the one that runs when the popover opens — is suppressed for `minInterval` (30 minutes) even though no release was ever seen.

Settings → **Check now** bypasses this (`minInterval: 0`), so the bug only shows up on the automatic path.

## Reproduction

1. Be on a build older than the latest GitHub release.
2. Force the latest-release request to fail (no network, or block `api.github.com`).
3. Open the menu-bar popover (triggers `updater.check()` with the default 30-minute interval).
4. Restore network and open the popover again within 30 minutes.

**Expected:** the failed attempt does not count; the next open retries and can show the banner.

**Actual:** the second open returns immediately because `lastChecked` was set on the failed attempt, so the banner stays hidden until the cooldown expires.

## Root cause

```swift
if let last = lastChecked, clock().timeIntervalSince(last) < minInterval { return }
lastChecked = clock()   // stamped before the request
guard let (data, resp) = try? await URLSession.shared.data(...) else { return }
```

A failed `guard` still leaves the stamp in place.

## Fix direction

Stamp `lastChecked` only after a trusted release payload is accepted. Keep an in-flight guard so overlapping popover opens do not stack concurrent GitHub calls once the early stamp is gone.

Borisserz:
## Summary

Fixes #320.

- `UpdateChecker.check()` stamped `lastChecked` before the GitHub request finished. A failed lookup (offline, timeout, bad payload) still started the 30-minute cooldown, so reopening the popover stayed silent even though no release was seen.
- Stamp the cooldown only after a trusted `https://github.com` release URL is accepted. Inject a fetch double in tests so the suite does not hit the network.
- Add an in-flight guard so overlapping popover opens do not stack concurrent requests once the early stamp is gone.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [ ] Other:

## Checklist

- [x] `swift build` and `swift test` pass locally (1081 tests, 13 skipped, 0 failures)

## Test plan

- [x] `testFailedCheckDoesNotStartTheCooldown` fails when `lastChecked` is stamped before fetch, then passes
- [x] `testSuccessfulCheckStartsTheCooldownAndAppliesTheRelease` passes
- [x] `testRejectedReleaseUrlDoesNotStartTheCooldown` passes
- [x] `swift build` and `swift test` locally: 1081 passed, 13 skipped, 0 failed
- [ ] CI `swift test`


chattymin (changes requested):
Please validate and normalize the release tag before recording . A malformed or otherwise unsupported tag must be treated as a failed check so the next attempt is not blocked for 30 minutes. Please add regression coverage for malformed and unsupported prerelease tags, and rebase onto the current main branch.

chattymin:
Clarification: the timestamp that must only be recorded after successful tag validation is lastChecked.

